PR #2, opened 2020-04-17
D8 un4

DuttonMa:
Multi site setup for Platform.sh.

Sets up three example sites and adds instructions to the README for adding further sites.

Note that these changes will break the local Lando installation procedure, more work to come for that in a later PR.

Note that the 'files' directory for each individual site lives at web/files/<site name> rather than the more standard sites/<site name>/files location. This was Crell's idea in an article that he wrote and it makes the config easier in .platform.app.yaml, at first I didn't like it but then decided that it wasn't a bad solution. We could change it so that we use the more standard sites/<site name>/files location but this would involve an extra step when adding a new site, as you'd have to add an extra 'mount' each time to .platform.app.yaml.

Note also that when setting up a site like uregni.gov.uk, the sites subdirectory needs to be sites/uregni which is again not what you would expect. It becomes way too complicated to configure platform if you use the more standard sites/uregni.gov.uk directory as the 'relationships' in .platform.app.yaml must be alpha names and do not work with any special characters (like full stops).


config/sync/core.extension.yml
```diff
@@ -89,14 +89,14 @@ module:
   menu_link_content: 1
   pathauto: 1
   views: 10
-  minimal: 1000
+  origins: 1000
```
johangant, line 92: This causes a CI break due to an outstanding issue in core right now. Leave this in, I think I need to detect/rewrite it in the CI job 👍

web/sites/fictcommission/settings.php
```diff
@@ -0,0 +1,89 @@
+<?php
```
johangant, line 1: These look the same across the site directories. Perhaps we can have either a symlink or an include statement in these files to avoid repetition?

web/sites/fictcommission/settings.php
```diff
@@ -0,0 +1,89 @@
...
+// Site hash salt.
+$settings['hash_salt'] = getenv('HASH_SALT');
+
+// Temp directory.
+$settings["file_temp_path"] = getenv('FILE_TEMP_PATH') ?? '/tmp';
+
+// Set config split environment; environment specific values is set near the end of this file.
+$config['config_split.config_split.local']['status'] = FALSE;
+$config['config_split.config_split.development']['status'] = FALSE;
+$config['config_split.config_split.production']['status'] = FALSE;
```
johangant, line 26: Yeah I don't know how nicely config split will behave with multisite or how it needs to be structured. Some extra research might be needed here :)

web/sites/fictcommission/settings.php
```diff
@@ -0,0 +1,89 @@
...
+      include $app_root . '/' . $site_path . '/../settings.development.php';
+  }
+}
+
+$settings['simple_environment_indicator'] = sprintf('%s %s', $env_colour, $env_name);
+
+// Set up a config sync directory.
+//
+// This is defined inside the read-only "config" directory, deployed via Git.
+$settings['config_sync_directory'] = '../config/sync/' . $subsite_id;
```
johangant, line 84: `'../config/sync'` has a corresponding environment variable we can use. But agreed, the subsite id will need to remain dynamic

web/sites/fictcommission/settings.php
DuttonMa: Good point, that's done now

PR #7, opened 2020-04-24
D8 un 5

DuttonMa:
Set up config split for the three example multi sites.

This looks impossible to review, given that there are over a thousand changed files ! Probably makes more sense to summarise my changes.

For each of the multi sites, I have created a config directory e.g. config/uregni.
Within this site directory there is a base 'config' directory (so config/uregni/config for uregni) and the sites/site.settings.php file sets the 'config_sync_directory' to this.
Following the lead from NIDirect, alongside this config/uregni/config' directory there are three other directories for 'local', development' and 'production' e.g. config/uregni/local. The directory settings on the config split admin screen select the appropriate one of these.

That's it really, the 'local' settings have been set up for each of the three example sites to install the modules that you would expect i.e. devel, easy_install, views_ui etc.


config/fictcommission/config/core.entity_form_display.node.article.default.yml
```diff
@@ -0,0 +1,105 @@
+uuid: e1104e52-0dbb-48dc-a72a-8837caf41035
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.node.article.body
+    - field.field.node.article.field_enable_toc
```
omahm, line 7: We renamed this field in NIDirect to _toc_enable, I'd suggest we do the same here.

johangant:
Needs a small adjustment to the CI script as otherwise the unit kernel tests fail on a sed command. I'll sort + push up to avoid too much context switching...

config/fictcommission/config/core.entity_form_display.node.article.default.yml
DuttonMa: Good point, have made the following changes:
- Field name changed in Origins Profile
- Origins Profile release completed
- Composer updated in this repo to include latest Origins Profile release
- Sites installed and config re-exported

PR #9, opened 2020-05-06
migrate upgrade settings added and migrate upgrade v3.2

Chris7791:


.lando.yml
```diff
@@ -27,6 +27,13 @@ services:
       user: drupal8
       password: drupal8
       database: drupal8
+  uregni7:
```
DuttonMa, line 30: We'll also need to add 'liofa7' and 'fictcommission7' in this Yaml file, and instructions in the README for future sites, but that can come in a later PR

PR #14, opened 2020-05-12
config module for uregni

Chris7791:


Chris7791:
Second commit merging Adam's spelling commit into feature branch

PR #15, opened 2020-05-13
D8UN-10 & D8UN-11 Basic page content migration

ooneill547:


config/uregni/config/core.entity_form_display.node.page.default.yml
```diff
@@ -25,48 +29,63 @@ content:
...
+  field_enable_toc:
+    weight: 2
+    type: boolean_checkbox
+    settings:
+      display_label: true
+    third_party_settings: {  }
+    region: content
+  field_page_top_image:
+    weight: 1
+    type: image_image
```
johangant, line 45: I'll flag this field type as something that will probably change in future by changing to a media entity reference field type. It may well be fine to leave as an image field for getting the basic page migration structure in place but if you're planning to use Drupal's media library functionality then later changes will be needed. See NIDirect for some good examples, in any case.

PR #27, opened 2020-05-20
Changing Postal address to to use a Text plain long field

soda736:


johangant:
@soda736 this looks okay to me, cheers. I might have missed some context around the need to move from an address field (structured) to a plain text field (unstructured). My guess would be... you don't need the extra complexity or structure. If that's the case, then this is good-to-go.

PR #33, opened 2020-05-22
D8 un 6

DuttonMa:
Solr multi site config for Lando and Platform.sh.
Instructions for adding Solr to new sites added to README.

There are an awful lot of files in this PR for the following reasons:
- I added quite a few Solr config files under .platform/solr_config (just copied over from NIDirect to give us a start point)
- when you install the 'search_api' and 'search_api_solr' modules they bring a lot of their own config and you also have to install the language module, so new fields are added to all content types


composer.json
```diff
@@ -97,7 +97,7 @@
         "harvesthq/chosen": "^1.8",
         "oomphinc/composer-installers-extender": "^1.1",
         "platformsh/config-reader": "^2.3",
-        "symfony/event-dispatcher": "4.3.11 as 3.4.35",
+        "symfony/event-dispatcher": "4.3.4 as 3.4.35",
```
johangant, line 100: Any reason this has dropped a minor point release? Probably of no consequence but struck me as unexpected.

config/uregni/config/language.mappings.yml
```diff
@@ -0,0 +1,13 @@
+map:
```
johangant, line 1: Is this config relevant to your PR or just stuff that needs to be part of the build anyhow?

composer.json
DuttonMa: This is an intentional upgrade to the Symfony Event Dispatcher - which is required for implementing Solr on Platform.sh (as described at https://docs.platform.sh/frameworks/drupal8/solr.html)

config/uregni/config/language.mappings.yml
DuttonMa: In order to install the search_api and search_api_solr modules I have to install the language module first, so I think that this is just a consequence of that.

PR #36, opened 2020-05-27
D8 un 38 theme news

soda736:


web/sites/uregni/themes/uregni_theme/templates/field/field--field-published-date.html.twig
```diff
@@ -0,0 +1,63 @@
...
+#}
+{% set attributes = attributes.addClass('published-date') %}
+{%
+  set title_classes = [
+  label_display == 'visually_hidden' ? 'visually-hidden',
+]
+%}
+
+{% if label_hidden %}
+  <strong>
```
David-Corbett, line 47: do you need the strong tag? If so, should it not sit inside the `<p>` tag

PR #46, opened 2020-06-08
D8 un 69

DuttonMa:
Install origins_workflow module and configure workflow and audit settings for Uregni


.platform.app.yaml
```diff
@@ -58,12 +58,21 @@ hooks:
   # The deploy hook runs after your application has been deployed and started.
   deploy: |
     set -e
-    for site in uregni liofa fictcommission; do
-      # Rebuild caches, run db-updates and import config.
-      drush -l $site -y cache-rebuild
-      drush -l $site -y updatedb
-      drush -l $site -y config-import
-    ; done
+
```
DuttonMa, line 61: Previous Yaml loop was not working, so replaced with the code below

config/uregni/config/system.site.yml
```diff
@@ -11,4 +11,4 @@ langcode: en
 default_langcode: en
 _core:
   default_config_hash: IqFDJOUXgoqYSm39bzsb0BVMDSkw3b4-KD4YSUgMw7U
-uuid: fe70d927-1fb3-47c8-b948-340813a132ab
+uuid: 69ec746f-0c27-4037-980c-b03885b42eaf
```
DuttonMa, line 14: We need to change this to match the uuid on platform.sh, so that changes can be deployed

config/uregni/config/core.extension.yml
```diff
@@ -59,9 +60,8 @@ module:
   token_filter: 0
   toolbar: 0
   twig_tweak: 0
-  update: 0
   user: 0
-  views_ui: 0
```
DuttonMa, line 64: Views UI removed from here as config is handled by config split.

.platform.app.yaml
johangant: Just saw this: `Hooks are executed using the dash shell, not the bash shell used by normal SSH logins. In most cases that makes no difference but may impact some more involved scripts.` At https://docs.platform.sh/configuration/app/build.html#build-hook

While this loop should work in bash/zsh/sh I've not tried it in dash shell. Could be that it needs a syntax tweak?

DuttonMa: Thanks, will investigate - it is quite unsatisfying not having that code in a loop !

PR #48, opened 2020-06-11
D8 un 70

DuttonMa:
Simplify config split in Unity sites to 'local' or 'hosted' as all sites on Platform.sh will be configured in the same way.
Also made changed to 'simple_environment_indicator' so that different colours are applied to the admin themes.
Also fixed a bug in the Platform.sh deploy code.


config/liofa/config/config_split.config_split.hosted.yml
```diff
@@ -0,0 +1,15 @@
+uuid: 3e35e375-2447-419b-b42c-cf933637aa6b
+langcode: en
+status: false
+dependencies: {  }
+id: hosted
+label: hosted
+description: 'Use for any sites that are ''''hosted'''' and not running locally, whether development, QA or production.'
```
soda736, line 7: Double quotes that aren't in other split files.

web/sites/settings.platformsh.php
```diff
@@ -123,6 +123,3 @@
 
 // Set the deployment identifier, which is used by some Drupal cache systems.
 $settings['deployment_identifier'] = $settings['deployment_identifier'] ?? $platformsh->treeId;
-
-// For now, set all Platform.sh sites to use 'production' config.
-$config['config_split.config_split.production']['status'] = TRUE;
```
johangant, line 128: +1 for this. I would try to keep their settings file as close to what they provide as possible and add our own changes into other settings files such as settings.php. That way if they make any significant changes, we shouldn't run the risk of losing any localised changes.

config/liofa/config/config_split.config_split.hosted.yml
DuttonMa: Thanks Adam, good spot

PR #49, opened 2020-06-12
Modifying consultations revision migration

soda736:


soda736:
Modifying the consultations revision yml file as I did with the consultations migration to split consultation start/end date into 2 fields.

PR #50, opened 2020-06-15
D8 un 43 44 vacancy migration

soda736:
- Content type/form/display settings fo vacancy content type.

- Added the metatag field to vacancy.

- I added a Uregni long date format for the vacancy closing date.

### DIFF
--- a/config/uregni/config/core.date_format.uregni_long.yml
+++ b/config/uregni/config/core.date_format.uregni_long.yml
@@ -3,6 +3,6 @@ langcode: en
 status: true
 dependencies: {  }
 id: uregni_long
-label: 'Uregni long'
+label: 'Uregni long, date and time'
 locked: false
 pattern: 'd F Y - H:i'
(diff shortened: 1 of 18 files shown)


config/uregni/config/core.date_format.uregni_long.yml
```diff
@@ -0,0 +1,8 @@
+uuid: f1b27c31-e8d8-4939-9215-a21f0570e7b5
+langcode: en
+status: true
+dependencies: {  }
+id: uregni_long
+label: 'Uregni long'
```
johangant, line 6: Small nitpick: I'd try a date format label like 'Long, with time' instead of 'Uregni long'... the latter doesn't mean anything to someone who doesn't already know the context of the site.

config/uregni/config/field.field.node.vacancy.field_salary.yml
```diff
@@ -0,0 +1,19 @@
...
+entity_type: node
+bundle: vacancy
+label: Salary
+description: ''
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings: {  }
+field_type: string
```
johangant, line 19: String should be fine for general information but if you need to sort any vacancies by salary then this could become problematic. I doubt this would be a major problem but I thought I'd flag it as a possibility.

PR #60, opened 2020-06-17
Update origins release and module

DuttonMa:
- update Origins release to fix published filter bug on admin/content page
- correct spelling of 'unity_file_migrations' module
- set site homepage for Uregni in line with source site
- comment out Liofa and Fictcommission deploys for now


.platform.app.yaml
```diff
@@ -66,16 +66,16 @@ hooks:
     drush -l uregni -y config-import
 
     # Liofa Rebuild caches, run db-updates and import config.
-    cd /app/web/sites/liofa
-    drush -l liofa -y cache-rebuild
-    drush -l liofa -y updatedb
-    drush -l liofa -y config-import
+    #cd /app/web/sites/liofa
```
johangant, line 69: A general observation and certainly not urgent; I'd delete + commit changes such as this rather than store commented out shell commands. It keeps things cleaner and if you need them, there's always git log/history to review.

DuttonMa: You are quite correct and I agree with this general sentiment. 
In this case, though, I am about to fix these in the next PR so think it is OK to leave them in for now.

PR #61, opened 2020-06-17
Multi site deployment

DuttonMa:
- Improve .gitignore
- Change name of unity_file_migrations module to improve clarity


.gitignore
```diff
@@ -35,12 +35,14 @@ console/
 
 # Ignore raw SQL files, potentially dangerous.
 *.sql
+*.sql.gz
```
johangant, line 38: Worth having 👍

.gitignore
```diff
@@ -35,12 +35,14 @@ console/
...
 
 # Misc tooling files to ingore.
 .swp
 .vscode
 .DS_Store
 .talismanrc
+*.json
```
johangant, line 45: This feels like it might need to be directory specific. Or at least confirming that it doesn't cause any issues if someone uses a .json file in a module or such like.

web/modules/custom/unity_file_migrations/unity_file_migrations.info.yml
```diff
@@ -1,6 +1,6 @@
-name: 'File Migrations'
+name: 'Migrate UNITY'
```
johangant, line 1: Super-nitpick for consistency: is it UNITY or Unity?

.gitignore
DuttonMa: Good point, have decided not to ignore it anymore !

web/modules/custom/unity_file_migrations/unity_file_migrations.info.yml
DuttonMa: I think that UNITY is best as it aligns nicely with the 'Migrate UREGNI' module

PR #62, opened 2020-06-18
Db performance

omahm:


.lando.yml
```diff
@@ -104,3 +79,75 @@ tooling:
   drush:
     server: appserver
     cmd: drush -r /app/web
+  'db-export [file]': disabled
+  'db-import <file>': diabled
```
johangant, line 83: Typo?

.lando.yml
```diff
@@ -104,3 +79,75 @@ tooling:
...
+    service: database
+    description: Import file to database
+    cmd: "sh /app/.lando/scripts/import_database.sh"
+    user: root
+    options:
+      database:
+        passthrough: true
+        description: Database to import to
+        alias:
+          - d
```
johangant, line 104: This alias is too long! 😜

.lando.yml
omahm: Shoulda went to Specsavers. Good spot !

.lando/scripts/import_database.sh
```diff
@@ -0,0 +1,28 @@
...
+    ;;
+    *)
+          # unknown option
+    ;;
+esac
+done
+
+if [ -f "$file" ]; then
+    echo "Dropping ${database} database"
+    mysql -u root -h $DB_HOST  -D $database -e "DROP DATABASE IF EXISTS ${database};"
```
DuttonMa, line 22: As discussed, need to remove this line, otherwise will not work with dumps from Platform.sh (as they have no 'create database' at the start)

PR #73, opened 2020-06-23
D8 un 19 entity browser config

soda736:
- Install paragraphs and various geolocation modules to deal with entity browser maps.

- Recreated nidirect_media for unity.

- Added map embed button and applied it to CKeditor text formats.


config/uregni/config/geolocation_google_maps.settings.yml
```diff
@@ -0,0 +1,4 @@
+google_map_api_key: ''
```
johangant, line 1: 👍  this should be sourced from/set environment variables as it's potentially sensitive. One to watch out for when exporting config locally

web/modules/custom/unity_media/src/Form/UnityEntityEmbedDialog.php
```diff
@@ -0,0 +1,37 @@
...
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'entity_embed_dialog';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, EditorInterface $editor = NULL, EmbedButtonInterface $embed_button = NULL) {
```
johangant, line 26: Goodness this brings back memories. Given that NIDirect and Unity share this, perhaps we should fold this into an origins_media module at some stage and incorporate to both sites?

soda736:
Hi @johangant, well I literally copied your module and work from NIDirect :) so yes everything should be more or less the same. I think that would be a good idea. Maybe need to discuss with the team?

PR #83, opened 2020-07-01
D8UN-53 Implement News search

ooneill547:
I took Adam's uregni_publications module and made it unity_search_pages. There is probably still a good bit of work to make it more universal for all unity sites.
Created the news search view and date published facet.

### DIFF
--- a/web/modules/custom/unity_search_pages/unity_search_pages.module
+++ b/web/modules/custom/unity_search_pages/unity_search_pages.module
@@ -2,7 +2,7 @@
 
 /**
  * @file
- * Custom code for publications.
+ * Custom code for search pages.
  */
 
 /**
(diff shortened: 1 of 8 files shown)


web/modules/custom/unity_search_pages/unity_search_pages.module
```diff
@@ -0,0 +1,51 @@
+<?php
+
+/**
+ * @file
+ * Custom code for publications.
```
soda736, line 5: Maybe change this to search pages.

ooneill547: good catch

web/modules/custom/unity_search_pages/unity_search_pages.module
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Implements hook_preprocess_views_view() for views_view.
+ * Adding a singular/pluralised result count in the header of the search pages.
+ *
+ * @RenderElement("html_tag");
+ */
+function unity_search_pages_preprocess_views_view(array &$variables) {
+  // Add view display ids to add result count to header of the search view.
+  $display_ids = ['publication_search_page', 'news_search_page'];
```
DuttonMa, line 16: This is a good idea, much better to re-use code

PR #84, opened 2020-07-02
D8 un 49 site search

soda736:


soda736:
I didn't even notice that. I was looking at Nidirect. I'll remove it for now as its not on the D7 site. Won't be hard to add it back in at a later stage if needed.

PR #90, opened 2020-07-08
D8 un 86 home news

soda736:


web/sites/uregni/themes/uregni_theme/templates/views/views-view--home-latest-news.html.twig
```diff
@@ -0,0 +1,41 @@
+{#
+/**
+ * @file
+ * Theme override for main view template.
```
johangant, line 4: Is it the main view? It looks like the latest news view template

soda736: Well spotted

PR #104, opened 2020-07-21
D8 un 94 footer blocks

soda736:
- Created the footer blocks and styled them.
- The footer logo block has an image in that hasn't exported but it needs some content in the body to export it so I have left it in. The image can be added into the dev environment at a later stage.


web/sites/uregni/themes/uregni_theme/src/scss/3_components/layout/footer.scss
```diff
@@ -0,0 +1,67 @@
...
+    font-style: normal;
+    margin-bottom: 2.4rem;
+  }
+
+  a {
+    color: $white;
+    text-decoration: underline;
+  }
+
+  input[type="submit"] {
```
omahm, line 29: I'd maybe suggest creating this as a reusable button component style.

PR #109, opened 2020-07-23
D8 un 61

Chris7791:
Functionality to find and replace publication embedded files with links to existing publication nodes, if they don't exist, create publication node and provide link to replace embedded file.

Works on News node type migrations currently.


web/modules/custom/unity_file_migrations/src/Plugin/migrate/process/DocumentEmbed.php
```diff
@@ -0,0 +1,87 @@
...
+        preg_match_all($embed_regex, $value['value'], $matches, PREG_SET_ORDER);
+
+        // Iterate these matches to find publications ( if exists ).
+        if (!empty($matches)) {
+
+            foreach($matches as $match) {
+
+                $query = \Drupal::entityQuery('node')
+                    ->condition('type', 'publication_page')
+                    ->condition('body', basename($match[1], "."), 'CONTAINS');
```
DuttonMa, line 53: Would be good to know that $match is actually an array and that $match[1] exists before you run this entity query - best to be defensive where possible.

web/modules/custom/unity_file_migrations/src/Plugin/migrate/process/DocumentEmbed.php
```diff
@@ -0,0 +1,87 @@
...
+                        'entity.node.canonical',
+                        ['node' => 2593],
+                        ['attributes' => ['rel' => 'nofollow', 'class' => 'publication_link']]
+                    )->toString();
+                }
+
+                // Then replace these links to publication link ( or publication page if not exists -- for now ).
+                //            $value = preg_replace($embed_regex, $publications_link, $value);
+
+                $value = str_replace($match[0], $publications_link, $value); // Str replace for multiple matches within body value
```
DuttonMa, line 80: Would be good to know that $match is actually an array and that $match[0] exists before you run this entity query - best to be defensive where possible.

web/modules/custom/unity_file_migrations/src/Plugin/migrate/process/DocumentEmbed.php
```diff
@@ -0,0 +1,87 @@
...
+                        ['node' => $lastNode],
+                        ['attributes' => ['rel' => 'nofollow', 'class' => 'publication_link']]
+                    )->toString();
+                } else {
+                    // Else, just create route to default publication page
+                    // Publication Page link used for replacing embedded links.
+                    $publications_link = Link::createFromRoute(
+                        'Default Publication Page',
+                        'entity.node.canonical',
+                        ['node' => 2593],
```
DuttonMa, line 72: I know that we discussed this and decided that it was OK to leave it hard coded like this for now - but it might be a good idea to add a comment to explain that

DuttonMa (approved):
Looks good to me now

DuttonMa (approved):
Nice work

PR #113, opened 2020-07-27
Fixing bugs deployment

soda736:


config/uregni/config/system.site.yml
```diff
@@ -4,7 +4,7 @@ slogan: ''
 page:
   403: ''
   404: ''
-  front: /home
+  front: /node/1
```
omahm, line 7: Should have spotted this before but Drupal has an issue using a path alias as the homepage.

PR #114, opened 2020-07-28
D8 un 99 theme searches

soda736:


config/uregni/config/views.view.news_search.yml
```diff
@@ -49,7 +49,7 @@ display:
       pager:
         type: full
         options:
-          items_per_page: 10
+          items_per_page: 20
```
ooneill547, line 52: The live site only has 10 items per page, has this been decided to change?

web/sites/uregni/themes/uregni_theme/templates/layout/page--news.html.twig
```diff
@@ -87,50 +87,22 @@
 
 <div id="container">
   <div class="container-wrap">
-    <div class="content-sidebar-wrap">
+    <div class="content-sidebar-wrap search-page">
```
ooneill547, line 90: Are new twig templates necessary for these pages? Would it be better to add code to add this class where appropriate?

soda736: I had a look into this but also need these twig templates to place the page title block. Can't find another way to do it. I added in a preprocess to add the class and removed the templates but lost the page titles. I think because the basic page hides the page title in block config it affects the views pages page titles.

PR #116, opened 2020-07-30
D8UN-108 Change text formats of content types

ooneill547:
Change text formats of content types and disable old D7 software responsive text formats


config/uregni/config/filter.format.filtered_html_with_tokens.yml
```diff
@@ -1,6 +1,6 @@
 uuid: 8baf6ed9-b0e8-4f51-9aaa-7cedd8943634
 langcode: en
-status: true
```
johangant, line 3: Is there any advantage to keeping these in config, but not active? I'd argue if they're legacy filters pulled over by a migration then they're better off dropped and Unity follows the config set in place with NIDirect.

Worth chatting with the team to get a decision on, either way 👍

composer.json
```diff
@@ -33,7 +33,7 @@
         "composer/installers": "^1.2",
         "cweagans/composer-patches": "^1.6.5",
         "dof-dss/nicsdru_origins_modules": "^0.4",
-        "dof-dss/nicsdru_unity_theme": "^0.1.1",
+        "dof-dss/nicsdru_unity_theme": "D8UN-108-dev",
```
DuttonMa, line 36: Think you should probably wait for a theme release rather than using a dev branch here

PR #120, opened 2020-08-04
Changing image fields to an entity reference field

soda736:
- Changing the image field to use an entity reference media field to match NIDirect.
- Updated displays to include new entity reference media fields.
- D8UN-98-responsive-images needs approval before image styles can be updated properly.
- Updated nicsdru_unity_theme package.
- Added in lingering config files, seem to just be updates to structure i.e. 1 - true, 0 - false.

### DIFF
--- a/config/uregni/config/core.entity_view_display.node.news.teaser.yml
+++ b/config/uregni/config/core.entity_view_display.node.news.teaser.yml
@@ -21,6 +21,15 @@ targetEntityType: node
 bundle: news
 mode: teaser
 content:
+  field_photo:
+    type: entity_reference_entity_view
+    weight: 0
+    region: content
+    label: hidden
+    settings:
+      view_mode: default
+      link: false
+    third_party_settings: {  }
   field_published_date:
     type: datetime_default
     weight: 1
@@ -42,7 +51,6 @@ hidden:
   body: true
   content_moderation_control: true
   field_meta_tags: true
-  field_photo: true
   field_summary: true
   langcode: true
   links: true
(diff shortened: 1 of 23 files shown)


config/uregni/config/core.entity_view_display.node.news.teaser.yml
```diff
@@ -24,13 +22,13 @@ bundle: news
...
   field_photo:
-    type: image
+    type: entity_reference_entity_view
     weight: 0
     region: content
     label: hidden
     settings:
-      image_style: nigov_landscape_xxl_960x640_x1
-      image_link: content
+      view_mode: default
```
johangant, line 30: We've tried to avoid using the default view mode on NIDirect as much as possible, especially for media, because sometimes new contrib modules can add elements to it which could create unwanted new things appearing when the entity renders.

config/uregni/config/field.storage.node.field_publication_image.yml
```diff
@@ -1,23 +1,17 @@
-uuid: b76f8830-ae86-4bb3-b1a4-d0179bfaa82d
-langcode: und
+uuid: df1a17ec-f9d5-4598-bd71-a6258823658d
+langcode: en
```
johangant, line 2: I've seen this set inconsistently on export in NIDirect. Probably not a big deal but it's unusual to have a langcode for a site that hasn't got language negotiation or detection enabled.

config/uregni/config/core.entity_view_display.node.news.teaser.yml
soda736: There are currently no view modes for responsive images set up until D8UN-98-responsive-images is approved. The view mode will be amended once that is merged in.

PR #128, opened 2020-08-12
Origins release

DuttonMa:
Update to latest version of origins modules


ooneill547:
Looks okay, but do we not need the composer.json also?

johangant:
Just spotted the GitGuardian checks picking up the Google Maps API key. It might be my old key in which case I'll create a new one or secure by hostname pattern... so this is okay to leave in place in this instance.

More broadly, it'll need some thought about how to store environment variables on Lando config without exposing them on a public Github repo.

They're generally fairly common in nature, so perhaps a separate private repo + pull in as part of the provisioning process might be a quick win.

johangant:
> Just spotted the GitGuardian checks picking up the Google Maps API key. It might be my old key in which case I'll create a new one or secure by hostname pattern... so this is okay to leave in place in this instance.
> 
> More broadly, it'll need some thought about how to store environment variables on Lando config without exposing them on a public Github repo.
> 
> They're generally fairly common in nature, so perhaps a separate private repo + pull in as part of the provisioning process might be a quick win.

API key now secured by hostname 👍

PR #131, opened 2020-08-17
D8 un 114 breadcrumbs

soda736:


web/sites/uregni/modules/custom/uregni_breadcrumbs/src/BoardBreadcrumb.php
```diff
@@ -0,0 +1,118 @@
...
+  public function applies(RouteMatchInterface $route_match) {
+    $match = FALSE;
+    $route_name = $route_match->getRouteName();
+    if ($route_name == 'entity.node.canonical') {
+      $this->node = $route_match->getParameter('node');
+      if ($this->node instanceof NodeInterface == FALSE) {
+        $this->node = $this->entityTypeManager->getStorage('node')->load($this->node);
+      }
+      if (!empty($this->node)) {
+        $match = $this->node->bundle() == 'staff_member';
```
DuttonMa, line 95: Personally, I don't like this syntax very much - I'm surprised that phpcs doesn't choke on it ! I would prefer to see at least some brackets here to make it clearer.

composer.json
```diff
@@ -65,6 +65,7 @@
         "drupal/linkit": "^4.3",
         "drupal/media_browser_responsive_images": ">1.x-dev",
         "drupal/media_entity_browser": "^2.0",
+        "drupal/menu_breadcrumb": "^1.13",
```
DuttonMa, line 68: Do we really need this contrib module ? 
Not sure what it is adding as you seem to have custom coded the breadcrumbs nicely in your classes below?

soda736: Discussed with Martin and Johan and have decided to leave it in for now.

web/sites/uregni/modules/custom/uregni_breadcrumbs/src/BoardBreadcrumb.php
soda736: Ran phpcs locally and got no errors.

johangant:
You'll want https://github.com/dof-dss/nidirect-drupal/commit/472044fa647bf3f307ea2366b9fc82ed0b6faeba in the codebase to help fix the CI job for kernel tests.

Regarding static analysis checks: you _can_ wire in phpcs to PHPStorm/other code editors and for some that's handy. In my experience to date, I tend to find it results in slightly inconsistent results because the configuration often drifts or is slightly out of sync with a CI service. If you want a preview of what CircleCI will tell you, run `lando phpcs` and - providing it's configured in the same manner as Circle CI - it'll chug away and review your codebase before you push your code.

PR #135, opened 2020-08-19
Reconfigure the uregni theme scss files to a more simple structure

ooneill547:
Remove directories in uregni theme and keep all of the scss code in a limited number of files


omahm (approved):
If there is very little styling it makes sense not to have it spread across multiple files. 👍

PR #136, opened 2020-08-20
D8 un 117 contact us

soda736:
Enabling Webform and related modules.
Deleting the existing webform content type and webform node module could no be installed without doing this.
Created a contact us webform.


omahm (approved):
Looks grand, could probably delete the default contact webform in future.

PR #140, opened 2020-08-24
D8 un 117 contact us

soda736:


config/uregni/config/core.extension.yml
```diff
@@ -48,6 +47,7 @@ module:
   history: 0
   image: 0
   inline_entity_form: 0
+  inline_form_errors: 0
```
johangant, line 50: Is that this module? https://www.drupal.org/project/ife 

If so, that doesn't seem to have a D8 release?

johangant (commented):
Looks sensible but I'm unclear where the inline_form_errors module is being sourced from.

config/uregni/config/core.extension.yml
soda736: It seems to come from core, it is enabled in NIDirect also.
https://www.drupal.org/docs/8/core/modules/inline-form-errors

PR #141, opened 2020-08-24
Update file paths uregni

ooneill547:
Update path to ckeditor.css and remove uregni specific styles library link


soda736 (approved):
Looks good

PR #143, opened 2020-08-26
Making footer region more semantic

soda736:


johangant (approved):
Looks good. One thought: there's a lot of repetition in there, have you seen https://twig.symfony.com/doc/2.x/tags/include.html before? It might be a nice enhancement at a later point.

PR #147, opened 2020-09-04
D8UN-127 Create field for homepage title in unity sites

ooneill547:
I have created a module to add two fields to the basic site settings form in unity sites. The first field is a textfield, frontpage_title, the second is a checkbox, frontpage_show_title, to determine whether the title will be visually rendered or not.

A line has also been added to the page.html.twig template to add the title in a h1 tag.


web/modules/custom/unity_frontpage/src/Form/ExtendedSiteInformationForm.php
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Add frontpage_title field to basic site settings front_page form.
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $site_config = $this->config('system.site');
+    $form = parent::buildForm($form, $form_state);
+    $form['front_page']['frontpage_title'] = [
+      '#type' => 'textfield',
+      '#title' => t('Front page title'),
+      '#default_value' => $site_config->get('frontpage_title') ?: 'Welcome to ' .
```
johangant, line 24: Although it's likely Drupal will escape anything malicious, it's worth noting that whenever you render something from the database or config to the client device you should be mindful of the possibility of unescaped user input rendering something that could act as a potential XSS or CSRF vulnerability. This is especially true when concatenating strings together.

To improve this, take a look at https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Render%21FormattableMarkup.php/class/FormattableMarkup/8.2.x : this allows you to safely mix in variables to strings. It's very commonly used in core and generally very readable, too :)

web/modules/custom/unity_frontpage/unity_frontpage.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'Unity Frontpage'
+type: module
+description: 'Module to add a front page title field to Unity sites'
+core: 8.x
```
johangant, line 4: I think we'll want to introduce 9 as an option here now. @omahm can point you at an example from NIDirect as he's been introducing these changes very recently.

web/modules/custom/unity_frontpage/unity_frontpage.install
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Implements hook_install().
+ *
+ * Sets frontpage_title value as default when module installed.
+ */
+function unity_frontpage_install() {
+  $config_factory = \Drupal::configFactory();
+  $config = $config_factory->getEditable('system.site');
+  $config->set('frontpage_title', 'Welcome to ' . $config->get('name') . ' site');
```
johangant, line 16: String concatenation here is likely fine as it'll be surfaced to an admin user or CLI output.

web/modules/custom/unity_frontpage/unity_frontpage.install
```diff
@@ -0,0 +1,30 @@
...
+  $config->save(TRUE);
+}
+
+/**
+ * Implements hook_uninstall().
+ *
+ * Unsets frontpage_title value when module is uninstalled.
+ */
+function unity_frontpage_uninstall() {
```
johangant, line 25: Great stuff, always good to see a module that tidies up after itself 👍

johangant (changes requested):
Really nice work! I have one concern around potentially unescaped output leaking to the page, but it's a very easy precaution to introduce.

johangant (approved):
LGTM 👍

PR #150, opened 2020-09-15
D8UN-139 Create and style vacancy view page

ooneill547:
Vacancy view page was created and styled to appear like the live site. 
I had to install the block visibility groups module to allow for more configuration of the sidemenu to display where appropriate.
Title and breadcrumb nav were also inserted into the vacancy view and vacancy node pages.


config/uregni/config/core.extension.yml
```diff
@@ -7,12 +7,14 @@ module:
...
   block: 0
   block_content: 0
+  block_visibility_groups: 0
   breakpoint: 0
   ckeditor: 0
   clientside_validation: 0
   clientside_validation_jquery: 0
   config: 0
   config_filter: 0
+  config_inspector: 0
```
DuttonMa, line 17: I think that 'config_inspector' here is probably a mistake and that you were just using it locally ?
If you do mean to include it in the build then you'll need to add it to composer too and make sure that it is only installed  on 'local' builds - not on platform.

DuttonMa (changes requested):
All looks good, just the one comment

config/uregni/config/core.extension.yml
ooneill547: I removed it but forgot to push core.extension when I pushed composer. I'll fix it now.

DuttonMa: All looks good now

DuttonMa (approved):
Looks good to me

PR #151, opened 2020-09-16
Removing duplicate colour variables and amending a a colour

soda736:
$green-dark was the same as $brand-secondary. So removing $green-dark.


DuttonMa (approved):
Looks good to me

PR #157, opened 2020-09-18
D8 un 137

soda736:


web/sites/uregni/themes/uregni_theme/src/scss/components.scss
```diff
@@ -313,7 +296,18 @@ h2 i {
...
+  @media screen and (max-width: $tab) {
+    padding-left: 1rem;
+  }
+
   .column {
+    padding-left: 0;
+
+    @media screen and (min-width: 556px) and (max-width: $tab) {
+      width: 50% !important;
+      float: left !important;
```
johangant, line 308: Hmm. Is there a better way of reducing the need for `!important` here? Maybe increase the specificity of selectors or even use an ID attribute to get this to take priority? `!important` will otherwise mess about with the priorities of layered CSS rules making longer-term maintenance much harder.

johangant (changes requested):
I'd encourage not using `!important` wherever possible

web/sites/uregni/themes/uregni_theme/src/scss/components.scss
soda736: I was using !important to override another !important. I have now amended the page.html.twig to match D7 and this gave me another ID attribute to use and reduce the need for !important declaration.

johangant (approved):
Cheers @soda736... always a pain when `!important` enters the scene but if that's what you're working against then any improvement's a good one.

PR #161, opened 2020-09-23
D8 un 152

soda736:
- Removing default classes on Consultations search and adding class to status.
- Adding more specificity to inputs as the existing style was overriding other inputs.
- Adding a border-bottom to all views rows on search pages on the site to keep them consistent.


DuttonMa (approved):
Looks good

omahm (approved):
LGTM

PR #162, opened 2020-09-23
D8 un 152

soda736:
Adding block visibility settings for breadcrumbs and page title.
Removing twig tweak and preprocess functionality for vacancy page.
Additional styling for search pages as the main search page didn't require same styles as other search pages.


omahm (approved):
LG

PR #164, opened 2020-09-24
Adding active link styling for sidebar nav

soda736:
D8UN-146 active-link styling sidebar nav.
D8UN-147 table th styling
Removing duplicate styles that are already in base.scss


omahm (approved):
LG

PR #167, opened 2020-10-02
D8UN-171 Only show h1 on page if a title has been set

ooneill547:


omahm (commented):
Not sure why you are checking the site config, can you explain? 
The render array should contain a '#title' element that could be checked for a value.

ooneill547:
> Not sure why you are checking the site config, can you explain?
> The render array should contain a '#title' element that could be checked for a value.

This is because I am pulling the value from a custom field that was added to the Basic Site Settings form through the Unity Frontpage module, it is not the 'Site name' value. This was the only method I could find to call this value.

DuttonMa (approved):
Looks good to me

PR #172, opened 2020-10-06
Update composer-drupal-optimizations to fix an error

ooneill547:
Getting the error `[ErrorException] Undefined index: extra` when trying to run `composer require` or `composer remove`. 

Based on information from https://github.com/composer/composer/issues/8047 it seems that this was caused by an issue with composer-drupal-optimizations module after a symfony update. 

The fix was found here: https://github.com/zaporylie/composer-drupal-optimizations/pull/22

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1973,9 +1973,6 @@
             },
             "type": "drupal-module",
             "extra": {
-                "branch-alias": {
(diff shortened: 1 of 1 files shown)


johangant (approved):
LGTM, thanks!

PR #182, opened 2020-10-14
D8UN-140

DuttonMa:
- Create a process filter for migration that scans the body field and changes any absolute links to Uregni sites to be relative links. This means that there will be no links on the dev site pointing to the live site, for example.
- Added this process filter to node migrations where appropriate
- Also deleted migration for 'project' content type as it is now obsolete and hasn't been implemented in D8

- After PR comments from Steve have now split this process filter into two separate filters and have parameterised them so that they can be used across all Unity sites


omahm (changes requested):
Good work. I'd make this a generic migration plugin and provide a parameter for the site specific stuff (url etc) so we can reuse it.

DuttonMa:
> Good work. I'd make this a generic migration plugin and provide a parameter for the site specific stuff (url etc) so we can reuse it.

Good idea, that's done now

omahm (approved):
Awesome work, much better that we can use this on any site 👍

PR #183, opened 2020-10-14
D8 un 183 diff

soda736:
I've added 2 diff field templates that use the same code (from classy). In NIDirect the diff templates extend core/themes/classy/templates/field/field.html.twig. I think classy is to be removed in Drupal 10 https://www.drupal.org/node/3103178. Not sure when this will be?

I've added another commit to use the extend functionality for now.


DuttonMa (approved):
Looks good to me, and I think that it is still fine to base templates on classy

PR #189, opened 2020-10-22
D8UN-185-file-bug

soda736:
Changing attachment field to entity reference


omahm (approved):
LGTM

DuttonMa:
As discussed, think we should just hold on to this change for now. It may be necessary to do this, but it would be better if we didn't have to. 
I'll re-run the Uregni migration this afternoon and we can re-assess then.

DuttonMa:
Actually, ignore that last comment, I think that we should go ahead and merge this now !

(It is not so much the migration of content that we need to worry about, more the fact that we want content editors to use the media entity browser for all attachments when creating future content.)

DuttonMa (approved):

DuttonMa:
Vacancy change looks good too, please go ahead and merge

PR #195, opened 2020-10-28
Changing heading colours to #111

soda736:


johangant:
I'm not quite sure whether 'text' is built-in value or a placeholder for something else. One for someone with more familiarity with this part of the stack to approve, I think!

ooneill547 (approved):
LG

PR #198, opened 2020-10-29
D8 un 221

DuttonMa:
Bug fix to 'publish_status' console command - node_field_revision table was often showing the wrong revision as 'published'.


omahm (approved):
LG

PR #203, opened 2020-10-30
D8 un 208

soda736:
Making a start on making search page exposed filters and forms look like dept-internet. This could change once the D8UN-208 ticket is fleshed out and some more needs done regarding wrapping meta list items.


DuttonMa (approved):
Looks good

PR #206, opened 2020-11-02
D8UN-224

DuttonMa:
Media document migration bug fix (this turned out to be the root cause of the problems that we were having with media documents in Consultations)


omahm (approved):

johangant (approved):
Amazing, did we have a funky field id when this was set up and was then later changed?

DuttonMa:
> Amazing, did we have a funky field id when this was set up and was then later changed?

I'm really not sure !

PR #209, opened 2020-11-04
D8UN-196

DuttonMa:
Finer grained permission checks on drafts views
- User must have 'quick publish' permission to publish from 'drafts' or 'all drafts' view
- If user does not have 'quick publish' permission but does have 'publish' then they can publish 'needs review' content but not 'draft' content

Relates to https://github.com/dof-dss/nicsdru_origins_modules/pull/108


omahm (approved):
LG

PR #210, opened 2020-11-04
Scheduler configuration for all content types

Chris7791:


config/uregni/config/node.type.staff_member.yml
```diff
@@ -4,11 +4,33 @@ status: true
 dependencies:
   module:
     - menu_ui
+    - origins_toc
+    - scheduler
 third_party_settings:
   menu_ui:
-    available_menus:
-      - main-menu
-    parent: 'main-menu:0:'
```
johangant, line 11: Was this intended? It's not scheduler/publish related but could be what you're looking for anyway... thought it prudent to check :)

johangant (approved):
Looks good to me, although one small question to follow up on around some menu config.

PR #222, opened 2020-11-10
D8 un 209

DuttonMa:
Create a new module that checks any body field that is saved and looks for absolute links.
If the absolute links match any of the patterns that have been added on the link checker admin page then they are changed to relative links.


omahm (approved):
Nice one, LGTM

PR #228, opened 2020-11-17
D8 un 222

DuttonMa:
Implement a post migration subscriber to update the publishing status of imported nodes.
Note that this code was already there but designed to be run manually as a console command after migration. 
Code has been refactored so that the same code is now run in both cases.


web/modules/custom/unity_file_migrations/src/EventSubscriber/PostMigrationSubscriber.php
```diff
@@ -0,0 +1,83 @@
...
+   * Handle post import migration event.
+   *
+   * @param \Drupal\migrate\Event\MigrateImportEvent $event
+   *   The import event object.
+   */
+  public function onMigratePostImport(MigrateImportEvent $event) {
+    $event_id = $event->getMigration()->getBaseId();
+    // Only process nodes, nothing else.
+    if (preg_match('/^upgrade_d7_node_/', $event_id)) {
+      $content_type = substr($event_id, 16);
```
johangant, line 78: Not sure if you need re-use here or not, but if so I'd pass the migrate id prefix as a variable and use the strlen to calculate the length for your substr call there. Otherwise, that could be a brittle point and fail in other contexts or if the migrate ids change for any reason.

johangant (approved):
Looks well put together. Is there any cross-over or common points between this and the new Drush powered commands for NIDirect?

omahm (approved):
Looks good to me. Johan has a point and there's also an argument for not writing out 'Clearing the caches...' to the shell, but it's personal choice.

PR #229, opened 2020-11-18
D8 un 250

DuttonMa:
Changes to make sure that media document titles are migrated across from D7 rather than just using the filename.
- Change migration to use title field
- Change vacancy attachment display to 'embed'
- Change to Uregni theme to add class to unordered list to pick up styling for vacancy attachments


omahm (approved):
LG

PR #233, opened 2020-11-24
D8UN-248

DuttonMa:
First implementation of nightly edge build for Unity (largely copied from the NIDirect version).
May require another nightly task to be added later to sync the data, but would like to release this job and then assess.


johangant (approved):
Looks good, thanks @DuttonMa. Platform deploy logs for the environment look sensible too 👍

PR #238, opened 2020-11-26
D8UN-179 Enable EU Cookie Compliance Module

ooneill547:


composer.json
```diff
@@ -91,7 +91,7 @@
         "drupal/pathauto": "^1.4",
         "drupal/redirect": "^1.3",
         "drupal/redis": "^1.4",
-        "drupal/scheduler": "^1.1",
+        "drupal/scheduler": "^1.3",
```
omahm, line 94: Ideally this should be in a different commit.

omahm (approved):
LG

PR #240, opened 2020-12-02
D8UN-236 Update ci config to purge and reindex solr index every night

ooneill547:


johangant (approved):
Looks good, nice work 😉

PR #241, opened 2020-12-02
D8UN-198

DuttonMa:
Created a Unity install profile that will be used to start off future Unity sites.
Decided to actually make the install profile a distribution, so that the user is not presented with a choice of install profiles to use.


.gitignore
```diff
@@ -9,7 +9,7 @@
 /web/libraries/
 /web/modules/migrate/
 /web/modules/origins
-/web/profiles/custom
+/web/profiles/custom/nicsdru_origins_profile
```
johangant, line 12: Just wondering, would  `/web/profiles/custom/*` or something to that effect cover this and any other renamings or other custom profiles?

composer.json
```diff
@@ -126,7 +126,6 @@
         "zaporylie/composer-drupal-optimizations": "^1.0"
     },
     "require-dev": {
-        "dof-dss/nicsdru_origins_profile": "^0.1.9",
```
johangant, line 129: This profile replaces the origins one listed here, right?

web/profiles/custom/nicsdru_unity_profile/config/install/system.performance.yml
```diff
@@ -0,0 +1,15 @@
+cache:
+  page:
+    max_age: 0
```
johangant, line 3: I'd set this at max by default to avoid development without internal or dynamic page cache: it often leads to confusion when work hits staging/prod environments.

web/profiles/custom/nicsdru_unity_profile/config/install/system.theme.yml
```diff
@@ -0,0 +1,2 @@
+admin: seven
```
johangant, line 1: Looks good, but not sure if NICS want to use adminimal as a base for admin themes? Might be one to discuss with the team.

web/profiles/custom/nicsdru_unity_profile/config/install/user.settings.yml
```diff
@@ -0,0 +1,16 @@
...
+notify:
+  cancel_confirm: true
+  password_reset: true
+  status_activated: true
+  status_blocked: false
+  status_canceled: false
+  register_admin_created: true
+  register_no_approval_required: true
+  register_pending_approval: true
+register: visitors_admin_approval
```
johangant, line 12: Might be the value needed (hard to tell from the values left by Drupal) but I think most NICS sites will want to only have admins create user accounts

johangant (approved):
I'm going to optimistically approve this PR... there are a couple of comments I've left, but TBH I don't think I can review the full set of YAML files with any degree of meaningfulness or quality :)

There's likely a discussion needed in future about how this profile might fit in with the original origins profile. Perhaps differences and common elements will become clearer with the departmental sites.

composer.json
DuttonMa: Yes, that's right - this profile is meant to be used for new Unity sites, so all of the Unity specific stuff will be enabled

web/profiles/custom/nicsdru_unity_profile/config/install/system.theme.yml
DuttonMa: Yes good point, but we've gone with Adminimal on Uregni so I think I'll leave this as is for now

web/profiles/custom/nicsdru_unity_profile/config/install/user.settings.yml
DuttonMa: Wow, that's an amazing catch, fixing it now

.gitignore
DuttonMa: Well actually we don't want to ignore that directory at all, as this new Unity install profile is in there now.
Have removed the reference to nicsdru_origins_profile though, as that was just for my testing and shouldn't have made it in to the PR

web/profiles/custom/nicsdru_unity_profile/config/install/system.performance.yml
DuttonMa: I'm not completely sure about that but have gone with your suggestion

PR #242, opened 2020-12-03
D8UN-236 Move sync into unity_edge_build_deploy

ooneill547:


johangant (approved):
So from what I've seen the platform redeploy command in the second scheduled workflow to sync data might well work better than what's on NIDirect, which sort of doubles up on itself with the composer tasks.

The diff itself here isn't terribly clear on the impact it has elsewhere in the pipelines, but from viewing the file in its entirety it _looks_ okay... probably a good idea to get it into action and assess how it responds overnight.

PR #243, opened 2020-12-03
D8 un 240

soda736:
New text format for publications to have only the image media type in the media browser.


config/uregni/config/filter.format.basic_html_publications.yml
```diff
@@ -0,0 +1,116 @@
...
+  config:
+    - core.entity_view_mode.media.embed
+  module:
+    - blazy
+    - editor
+    - entity_embed
+    - media
+    - noreferrer
+    - token_filter
+name: 'Basic HTML publications'
```
DuttonMa, line 14: So will this appear to the content editor as 'Basic HTML publications' in the dropdown at the bottom of the body field ? 
If so, would it be better just to display it as 'Basic HTML' as they don't need to know that it is different ?

DuttonMa (commented):
Looks great, just one question

config/uregni/config/filter.format.basic_html_publications.yml
soda736: I did try to call it Basic HTML but it said that names had to be unique, even though the machine name was different.

Content editors will not see a dropdown if they can only use one text format. So they will think that they are using Basic HTML anyway.

DuttonMa: OK, sounds good, now approved !

DuttonMa (approved):

PR #245, opened 2020-12-04
reinstate origins profile and re-organise profile directory

DuttonMa:
I had previously removed the Origins install profile as we should use the Unity install profile from now on.
The current Uregni site is built on the Origins profile, however, so to avoid errors it is necessary to reinstate it.
Have re-organised the profiles directory so that we can have one profile (Origins) being handled by composer and another (Unity) that is held in this repo.
Note that there are a huge number of Yaml files in this PR but nothing much has changed - they are just being moved from web/profiles/custom/nicsdru_unity_profile to web/profiles/unity/nicsdru_unity_profile


soda736 (approved):
LG

.circleci/config.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           at: ./
       - run:
           name: PHPCS static analysis
-          command: /home/circleci/project/phpcs.sh /home/circleci/project "/home/circleci/project/web/sites /home/circleci/project/web/modules/origins /home/circleci/project/web/modules/custom /home/circleci/project/web/themes/custom /home/circleci/project/web/profiles/custom"
+          command: /home/circleci/project/phpcs.sh /home/circleci/project "/home/circleci/project/web/sites /home/circleci/project/web/modules/origins /home/circleci/project/web/modules/custom /home/circleci/project/web/themes/custom /home/circleci/project/web/profiles/unity /home/circleci/project/web/profiles/origins"
```
soda736, line 84: There is a space after unity in the path.

PR #248, opened 2020-12-07
D8 un 254

DuttonMa:
- Make 'publish status' processing run only after revision migrations
- Add processing for nodes in 'needs review' moderation state when migrated


omahm (approved):
LG

soda736 (approved):
LG

PR #261, opened 2020-12-21
D8 un 248 part3

DuttonMa:
- Structural changes to Edge build, now syncs data rather than re-deploying
- Added steps to build Uregni theme

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -271,6 +271,25 @@ jobs:
             git add composer.*
             git commit -m "Set dof-dss packages to HEAD development for build"
 
+            # Need to re-build Uregni theme with any Unity changes.
+            cd web/sites/uregni/themes/uregni_theme
+            sed -i -e "s/\"nicsdru_unity_theme.*/\"nicsdru_unity_theme\": \"github:dof-dss\/nicsdru_unity_theme#development\",/" package.json
+
+            # Download and configure node and npm.
+            curl -sL https://deb.nodesource.com/setup_12.x | sudo bash -
+            sudo apt install nodejs
+            sudo apt install npm
+            sudo apt install automake
+            sudo apt install libjpeg-dev
+
+            # Build Uregni theme.
+            npm install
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -271,10 +271,27 @@ jobs:
...
+            # Download and configure node and npm.
+            curl -sL https://deb.nodesource.com/setup_12.x | sudo bash -
+            sudo apt install nodejs
+            #sudo apt install npm
+            sudo apt install automake
+            sudo apt install libjpeg-dev
+
+            # Build Uregni theme.
+            npm install
+            npm rebuild node-sass
```
johangant, line 287: Arguably this kind of stuff should be done/handled by a CI process, just like this, so we don't need to compile+commit CSS/JS into the canonical repository in GitHub. One for future discussion!

.circleci/config.yml
```diff
@@ -271,10 +271,27 @@ jobs:
             git add composer.*
             git commit -m "Set dof-dss packages to HEAD development for build"
 
-            # If edge build branch already exists, delete it before pushing the new one.
-            if git branch -rr | grep $EDGE_BUILD_BRANCH; then
-              git push origin :$EDGE_BUILD_BRANCH
-            fi
+            # Need to re-build Uregni theme with any Unity changes.
+            cd web/sites/uregni/themes/uregni_theme
+            sed -i -e "s/\"nicsdru_unity_theme.*/\"nicsdru_unity_theme\": \"github:dof-dss\/nicsdru_unity_theme#development\",/" package.json
```
johangant, line 276: Looks good. Probably not as fast because it will need to calculate the dependency tree and make changes to the file, but a more robust option might be to re-point the package reference here using a npm command?

Don't think this needs any active changes... just something to consider if this ever breaks in future.

johangant (approved):
_kicks the tyres_ LGTM 👍 

Some comments for prospective future enhancements.

PR #267, opened 2021-01-07
D8UN-242 Amend search page titles on search or facet selection

ooneill547:
Add a subtitle to search pages when a keyword has been searched or a facet selected.


web/modules/custom/unity_search_pages/unity_search_pages.module
```diff
@@ -57,3 +54,22 @@ function unity_search_pages_preprocess_views_view(array &$variables) {
...
+
+/**
+ * Implements hook_preprocess_page_title().
+ *
+ * Replaces search results title with a title and subtitle to enable styling.
+ */
+function unity_search_pages_preprocess_page_title(&$variables) {
+  $current_route = \Drupal::routeMatch()->getRouteName();
+  $title = $variables['title'];
+  $pattern = '/view.\w*search.\w*search_page/';
```
johangant, line 66: Not a required change at all, but if you've got a pattern that contains lots of path dividers (eg: `/`) that need escaping then the regex itself gets rather unreadable. In that instance you can switch the default pattern delimiters to something else. Eg: `/one\/two\/three/` could become `#one\/two\/three#'

web/modules/custom/unity_search_pages/src/Routing/SearchPagesRouteSubscriber.php
```diff
@@ -0,0 +1,30 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterRoutes(RouteCollection $collection) {
+    foreach ($collection->getIterator() as $route_name => $route_info) {
+      $pattern = '/view.\w*search.\w*search_page/';
+      if ((preg_match($pattern, $route_name)) || ($route_name == 'search.view')) {
+        if ($route = $collection->get($route_name)) {
+          $route->setDefault('route', $route_name);
+          $route->setDefault('_title_callback', '\Drupal\unity_search_pages\Controller\SearchPagesController::getTitle');
```
johangant, line 24: As an optimisation, once you find your required match/take an action here, you might want to return/break the parent loop to avoid iterating through the remainder of the collection which could be quite large.

johangant (approved):
Looks good to me. A few small comments which might be helpful for optimisations or future regex legibility.

web/modules/custom/unity_search_pages/src/Routing/SearchPagesRouteSubscriber.php
ooneill547: There will likely be multiple matches per site (each of the search pages) so I'm not sure it would be possible to implement that.

PR #276, opened 2021-01-21
D8UN-300

soda736:
Amending regex to only match divs or lis before the file link.


ooneill547 (commented):
Will this still target a link if it is at the start of an <li> in the body text?

soda736:
Yes, I think this will match any 2 or 3 letter lower case pattern inside <> tags. I don't want it to match <p> which would be 1 letter and every migrated file link I could see was inside either a <div> or <li>. Maybe there is a better way to target this though?

DuttonMa (approved):
Looks good to me, as far as I can see this regex will do just what you need it to

PR #279, opened 2021-01-22
new facet drilldown contrib module

DuttonMa:
As discussed in the team meeting, this is the new facet drilldown contrib module that has been written by Orla, Adam and me.

We do not intend to merge this PR, but it is a good way to ask people to review it and to collect any feedback. Any changes that result from comments on this PR will be pushed up to drupal.org.


web/modules/custom/date_facet_drilldown/date_facet_drilldown.module
```diff
@@ -0,0 +1,64 @@
...
+  // The purpose of this function is to correct the URL for the active year
+  // facet, as this does not work otherwise.
+  // Not ideal to do this in a preprocessor but the alternative was to
+  // implement an entire new facet URL processor as well as a route subscriber
+  // - this not only introduced a lot of code but involved copying route
+  // subscriber code over from facets_pretty_paths which did not feel good.
+  if (!empty($variables["facet"])) {
+    if ($procs = $variables['facet']->getProcessorConfigs()) {
+      // Check that our 'date_facet_drilldown' processor has been selected.
+      if (isset($procs['date_facet_drilldown'])) {
```
johangant, line 31: I usually recommend the use of `empty` over `isset` because the latter can evaluate to true if, say, a variable or array key is instantiated rather than checking whether it has a non-null value. I can see `empty` is used further up in the fucntion too for a similar check.

web/modules/custom/date_facet_drilldown/date_facet_drilldown.module
```diff
@@ -0,0 +1,64 @@
...
+ * Implements hook_preprocess_facets_item_list().
+ */
+function date_facet_drilldown_preprocess_facets_item_list(array &$variables) {
+  // The purpose of this function is to correct the URL for the active year
+  // facet, as this does not work otherwise.
+  // Not ideal to do this in a preprocessor but the alternative was to
+  // implement an entire new facet URL processor as well as a route subscriber
+  // - this not only introduced a lot of code but involved copying route
+  // subscriber code over from facets_pretty_paths which did not feel good.
+  if (!empty($variables["facet"])) {
```
johangant, line 28: A personal opinion here: six levels of conditional code is usually a flag for me to stop and consider whether I can refactor to combine simpler checks, provide some early 'short circuits' or quick exit options at the start of the function, or encapsulate complex parts into their own functions to make it easier to comprehend.

web/modules/custom/date_facet_drilldown/date_facet_drilldown.module
```diff
@@ -0,0 +1,64 @@
...
+          foreach ($variables['items'] as $key => $item) {
+            if (preg_match('/^\d{4}$/', $key)) {
+              // This must be a 4 digit year.
+              $current_route = \Drupal::routeMatch()->getRouteName();
+              $route_params = \Drupal::routeMatch()->getParameters();
+              $params = $route_params->getIterator()['facets_query'];
+              // Manipulate the parameters, removing the date facets.
+              $url_alias = $variables['facet']->getUrlAlias();
+              // Remove 'date/YYYY-MM'.
+              $params = preg_replace('|' . $url_alias . '\/\d{4}-\d{2}|', '', $params);
```
johangant, line 45: +1 for using a regex delimiter that isn't a forward slash! Makes it much easier to read when there are multiple, escaped slash characters in the pattern already.

johangant (approved):
LGTM. Plenty of parts to this. I could only really chime in on some parts that felt a bit complex to read/comprehend that might benefit from refactoring a little, and some other minor stuff.

web/modules/custom/date_facet_drilldown/src/Plugin/facets/query_type/SearchApiDateFacetDrilldown.php
```diff
@@ -0,0 +1,93 @@
...
+ *   id = "search_api_datefacetdrilldown",
+ *   label = @Translation("Date facet drilldown"),
+ * )
+ */
+class SearchApiDateFacetDrilldown extends QueryTypeRangeBase {
+
+  /**
+   * @inheritdoc
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
```
omahm, line 25: Shouldn't need this as all it's doing is calling the parent constructor.

omahm (changes requested):
Good job 👍  
There's a constructor in there that isn't really needed.

web/modules/custom/date_facet_drilldown/src/Plugin/facets/query_type/SearchApiDateFacetDrilldown.php
DuttonMa: Good point, removing that

web/modules/custom/date_facet_drilldown/date_facet_drilldown.module
DuttonMa: Have now broken out to a separate function

web/modules/custom/date_facet_drilldown/date_facet_drilldown.module
DuttonMa: Fair enough, have changed that

PR #285, opened 2021-01-27
D8 un 291

soda736:
Adding the https://www.drupal.org/project/menu_admin_per_menu contrib module. This module gives a permission for each menu on the site whereas the core config permission is access all menus which we don't want a supervisor to access. We only want them to access the main menu which this module allows us to do.


DuttonMa (approved):
Sounds like a good idea

PR #292, opened 2021-02-01
Install facet drilldown module

DuttonMa:
- install new facet drilldown contrib module that has been created by the Unity team
- configure consultation, publication and news search facets to use the new module


johangant (approved):
LGTM

PR #298, opened 2021-02-10
D8UN-310 Set up sites and export config

ooneill547:
Six new sites have been set up using installation profile and default config exported.


DuttonMa (approved):
Not really possible to review this comprehensively !
Have looked for a few gotchas and it all looks good to me as far as I can see.
I suggest that we merge it and see how it goes.

PR #300, opened 2021-02-12
D8UN-310 Update config split configuration

ooneill547:
The config split config was incorrect in the profile which caused unexpected behaviour. This has been changed and config exported again.


DuttonMa (approved):
Looks good to me

PR #335, opened 2021-03-11
D8 un 360 b

DuttonMa:
- Bug fix to Drupal console 
- Pull latest version of date_facet_drilldown module as it correctly handles active year facets when duplicated by the facets_summary module

### DIFF
--- a/.lando.base.yml
+++ b/.lando.base.yml
@@ -35,6 +35,8 @@ services:
     hogfrom:
       - appserver
 tooling:
+  drupal:
+    cmd: "/app/vendor/bin/drupal --root=/app/web"
   xdebug-on:
     service: appserver
     description: Enable xdebug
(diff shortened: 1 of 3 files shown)


.lando.base.yml
```diff
@@ -35,6 +35,8 @@ services:
     hogfrom:
       - appserver
 tooling:
+  drupal:
+    cmd: "/app/vendor/bin/drupal --root=/app/web"
```
DuttonMa, line 39: I know that the Drupal console is being phased out - but we still have some Drupal console commands in Unity so don't want it to go just yet !

soda736 (approved):
LG

omahm (approved):
LG, I thought lando had Drupal console still as part of the recipe

DuttonMa:
> LG, I thought lando had Drupal console still as part of the recipe

Yes it does, but it wasn't working - this is a bug fix that I put in so that 'lando drupal' works correctly

PR #337, opened 2021-03-12
D8UN-369

soda736:
Updating the regex for publication file links. After the latest database update some of the publications are now wrapped in a `<p>` so changing the 1st part to look for 1,3 letters rather than 2,3.


DuttonMa (approved):
Looks good to me

PR #338, opened 2021-03-15
Google map field

soda736:
Config for the google map field. This will be re-usable over all sites I imagine. Should this config also be added to either origins_media or unity install profile?


johangant:
If this is what we use on NIDirect too perhaps it's a good point to refactor both to bring the common elements into an origins module?

DuttonMa (commented):
Think you are right to bring this up Adam, probably best to hold off merging this whilst we decide ......

DuttonMa (commented):
I see that you haven't added a Google maps api key in config, which is quite correct, but we'll need to make sure that we have one up on the Platform sites

DuttonMa (approved):
Actually, I'm now thinking that you can go ahead and merge this. It won't hold up your hatecrimereviewni site and we can then centralise the config afterwards.

PR #339, opened 2021-03-15
Ckeditor settings

soda736:
Deleting D7 migrated filter formats.
Setting up D8 filter formats. This config could be added to an origins module or unity install profile as they should be the same across all codebases.


ooneill547 (commented):
I think some text format configuration is in the unity install profile, however, it needs to be updated

DuttonMa (approved):
Looks good, and good thought again about setting this up for all sites - I'll take a look

PR #343, opened 2021-03-15
D8UN-315 set up basic page

ooneill547:


soda736 (approved):
LG

config/odscni/config/core.extension.yml
```diff
@@ -86,6 +86,7 @@ module:
   toolbar: 0
   twig_tweak: 0
   unity_common: 0
+  unity_file_migrations: 0
```
DuttonMa, line 89: We shouldn't install any migration modules in config, as we don't want them to be installed on any Platform.sh sites, we just need to install them locally when we're working with migrations (see Uregni).

DuttonMa (changes requested):
Just one module to uninstall, otherwise looks great

config/odscni/config/core.extension.yml
ooneill547: This has now been removed

soda736 (approved):
Looks good

DuttonMa (approved):

PR #346, opened 2021-03-16
D8 un 374

DuttonMa:
Turn off Platform.sh HTTP Caching for Unity sites. 
It appears that the Platform.sh HTTP Cache does not work with Drupal cache tags and therefore should never be used for Drupal 8 sites. 
For high traffic sites we could implement Fastly, but otherwise we'll just have to rely on Drupal caching and Redis.


omahm:
> Turn off Platform.sh HTTP Caching for Unity sites.
> It appears that the Platform.sh HTTP Cache does not work with Drupal cache tags and therefore should never be used for Drupal 8 sites.
> For high traffic sites we could implement Fastly, but otherwise we'll just have to rely on Drupal caching and Redis.

Pretty surprised by that!

DuttonMa:
> > Turn off Platform.sh HTTP Caching for Unity sites.
> > It appears that the Platform.sh HTTP Cache does not work with Drupal cache tags and therefore should never be used for Drupal 8 sites.
> > For high traffic sites we could implement Fastly, but otherwise we'll just have to rely on Drupal caching and Redis.
> 
> Pretty surprised by that!

Well, it would perhaps have been more accurate for me to say that there is no way for Drupal to invalidate the Platform.sh HTTP Cache. As it says in Platform's own documentation (at https://docs.platform.sh/bestpractices/http-caching.html):

_"Every project includes a router instance, which includes optional HTTP caching. It is reasonably configurable and obeys HTTP cache directives, but does not support push-based clearing. "_

ooneill547 (approved):

PR #363, opened 2021-03-22
D8UN-318 Create ODSCNI subtheme

ooneill547:
The svg logo is to follow once received from Richard


web/sites/odscni/themes/odscni_theme/config/postcss.config.js
```diff
@@ -0,0 +1,21 @@
...
+  },
+  plugins: {
+    'autoprefixer': {
+      cascade: false,
+    },
+    'postcss-assets': {
+      cachebuster: true,
+      relative: true,
+      loadPaths: ['images'],
+      baseUrl: '/themes/custom/STARTERKIT/'
```
soda736, line 18: I think this needs changed to the name of the theme.

soda736: Looks good, there Iip-logo.png is in the images as well which we won't need.

ooneill547: Good catch! Thanks

soda736 (approved):
Looks good now

PR #375, opened 2021-03-25
News search

soda736:
Setting up the news search. It was decided to use Search API rather than Solr for this site as it is only 1 index and the site lifespan is limited.


ooneill547 (commented):
Looks good, however, not sure what the comment config is about? Is there a comment section on this site?

soda736:
I thought it was a dependency on Search API but it's not! I have uninstalled it and deleted all related config.

ooneill547 (approved):

PR #382, opened 2021-03-27
Homepage styling

soda736:
Adding a caption field that the homepage block needs for its image and a new font-style for the caption.


omahm (approved):
Looks grand.

PR #385, opened 2021-03-29
ODSC Install modules and update general config

ooneill547:
Installed:

- EU Cookie Compliance
- Scheduled Transitions
- Content lock

Removed:

- Scheduler
- Extra filter formats
- News content type


DuttonMa (approved):
All looks good to me

PR #386, opened 2021-03-30
D8 un 370

DuttonMa:
Updates to post migration processing to rectify problems found after recent Uregni migration.
This code now takes the more mainstream approach of updating nodes and revisions through the entity type manager rather than updating MySQL tables directly.


ooneill547 (approved):
Looks fine as far as I can tell

johangant (approved):
Good to see more pushed through the APIs. We discussed revision id drift due to the two-step update on Slack, but providing the APIs keep other referencing things in sync then this looks good to me.

PR #388, opened 2021-03-30
Create a module to configure the map embed button in CKEditor

ooneill547:
Based on discussion around updating the unity install profile, I have created a module to create the map embed button for CKEditor.


DuttonMa (approved):
Good idea

PR #389, opened 2021-03-30
ODSC config update

ooneill547:
Following changes made to the unity install profile, I noticed that there were modules missing or not configured appropriately on the ODSC site, including:

- media_libarary_edit
- ultimate_cron
- menu_admin_per_menu
- and those required to set up the map embed button in CKEditor


omahm (approved):
LG

PR #391, opened 2021-03-31
D8UN-385

soda736:
Adding some preprocessing from origins_media that we can't use because of references to nicsdru_origins_theme. It was advised to extract these and put them in a unity module until it can be resolved. This is part of the D8UN-385 PR in nicsdru_unity_theme.


web/modules/custom/unity_common/unity_common.module
```diff
@@ -38,3 +43,145 @@ function unity_common_form_alter(&$form, FormStateInterface $form_state, $form_i
...
+    if ($file instanceof FileInterface === FALSE) {
+      return;
+    }
+
+    $mimetype = $file->getMimeType();
+
+    // There shouldn't be an image file here, but just in case, screen for it as if
+    // it's genuine it likely has its own thumbnail so we would want to keep that.
+    if (preg_match('/jpeg|jpg|gif|png/', $mimetype)) {
+      return;
```
DuttonMa, line 124: Nice use of regex

DuttonMa (approved):
Nice work here.
I did wonder if this code should possibly be moved out to another module rather than going in to unity_common? 
But we did have a unity_media module before and it was deleted so reinstating that module name might cause some confusion. If we are sure that this functionality will be needed on every Unity site then it is fine to stay here.

web/modules/custom/unity_common/unity_common.module
soda736: I did think that too and same about unity_media so as it is used throughout all sites I thought unity_common was a good place.

This is all from origins_media so it's someone elses regex :)

PR #392, opened 2021-03-31
Config refinements

johangant:
Some small refinements:

* Better detection of Lando environment (it sets its own env var now)
* Add trusted host pattern for any Lando site domain (stops the warning on the status report page)
* Changed something in the README about loading the DB dump as it wasn't terribly clear it was an either/or and not site-id/db-name.
* Sorted out drupal-check versions and dependencies
* Added stage file proxy config via env var (safer than adding into YML file)


web/sites/site.settings.php
```diff
@@ -81,11 +81,9 @@
...
 }
-else {
-  // Lando config.
-  if (file_exists($app_root . '/' . $site_path . '/../settings.lando.php')) {
-    include $app_root . '/' . $site_path . '/../settings.lando.php';
-  }
+
+if (getenv('LANDO') && file_exists($app_root . '/' . $site_path . '/../settings.lando.php')) {
+  include $app_root . '/' . $site_path . '/../settings.lando.php';
 }
```
DuttonMa, line 87: I'm not sure that we needed to change this really, at least before (with the 'else') it was guaranteed to execute one of the options.
Are we confident that the 'LANDO' environment variable will be set ? If it isn't we could get some strange results.

web/sites/sites.php
```diff
@@ -63,10 +63,10 @@
...
-} else {
+}
+
+if (getenv('LANDO') && file_exists($app_root . '/sites/sites.lando.php')) {
   // Running in Lando locally, include appropriate sites file.
-  if (file_exists($app_root . '/sites/sites.lando.php')) {
-    include $app_root . '/sites/sites.lando.php';
-  }
+  include $app_root . '/sites/sites.lando.php';
 }
```
DuttonMa, line 72: Same comment as above really

README.md
```diff
@@ -25,7 +25,7 @@ Or, if available, you may also fetch the database and import this:
 
 `platform db:dump` (you'll need to select project, environment and required site e.g. 'uregni')
 
-`lando db-import -d <sitename/database> -f <downloaded sql file>` (where 'site name' may be 'uregni', 'liofa' etc)
+`lando db-import -d <sitename> -f <downloaded sql file>` (where 'sitename' may be 'uregni', 'liofa' etc)
```
DuttonMa, line 28: Yes, good spot

web/sites/settings.lando.php
```diff
@@ -35,6 +35,14 @@
 $settings["file_temp_path"] = getenv('FILE_TEMP_PATH') ?? '/tmp';
 $settings['file_private_path'] = getenv('FILE_PRIVATE_PATH');
 
+// Add trusted host pattern for Lando sites, in Lando config.
+$settings['trusted_host_patterns'][] = '^.+\.lndo\.site$';
+
```
DuttonMa, line 40: This is a good idea

DuttonMa (approved):
Overall I like these changes, just a slight reservation about the 'else's !

PR #393, opened 2021-04-01
D8 un 370 b

DuttonMa:
Stop the post migrate publish processing from running automatically after a revision migration finishes. This is because we only want to run this once after all migrations have finished, so it should be run manually from now on.


johangant (approved):
👍

PR #399, opened 2021-04-07
D8UN-403 Add and configure new unity sites

ooneill547:
Sorry, a pretty big PR.

Includes:
- Updates to unity install profile
- Set up for new sites (fiscal commission and fiscal council)
- Config export for fiscal commission site
- Config export for fiscal council site


johangant (approved):
I think the only helpful thing I can put here is LGTM, because it's far too big to review in detail.

In all fairness, it's not in production and what I have seen so far looks sensible so I think this a case of "suck it and see" / adjust as needed afterwards.

There's definitely a separate case to look at sharing more common config between sites to avoid repetition and reduce future maintenance overheads but that's outside of the scope of this PR.

PR #406, opened 2021-04-09
D8UN-403 Update fiscalcommissionni config

ooneill547:
Config for fiscal commission site has been re-exported after domain name change


DuttonMa (approved):
Looks good, shame that you had to do this but I suppose that it is good to stick to our conventions for all sites!

PR #424, opened 2021-04-14
Improve env var handling

johangant:
Firing up a new project will need people to:

- Copy the sample env file for Lando to use.
- Fill in the values. They can either be known, public values (eg: Lando) or for the sensitive ones they'll need access to the hosting environment to get those values.
- The .gitignore file now prohibits any `.env` files being committed: they're always used for environment specific config and can often hold sensitive keys in public repos committed by accident.


DuttonMa (approved):
Looks good

PR #425, opened 2021-04-14
D8UN-380

johangant:
- Drops talisman support (we can now rely on GitGuardian for this)
- Prefixes env vars with subsite_id (worked out by Drupal to assess which site to use in a multisite codebase) so we can manage variables for API keys across sites on the same codebase and with the same code.


omahm (approved):
Looks good

PR #442, opened 2021-04-20
D8UN-427 Fiscal Comm: Publication content type

ooneill547:
- Update the document media display
- Add 'Basic HTML publications' filter format and apply it to publication body
- Add url pattern for content type
- Changes to content type to align with ticket


soda736:
I don't think the Basic HTML publications format is needed for this site as this content type seems to have an attachment field. The Basic HTML publications format was only added to UREGNI to deal with migrated publications. I think it should be the regular Basic HTML format used.

DuttonMa:
Just a reminder that we shouldn't merge anything in to development until we have finished the D9 upgrade - probably Thursday morning earliest

DuttonMa (approved):
Looks good to me, hopefully you can get those conflicts sorted out fairly easily

PR #458, opened 2021-04-22
Amendments to install profile

soda736:
Made some changes to the install profile:
All content types in the install profile are now added to scheduled transitions.
Supervisor has access to them all.
I can't get the cookie popup_info or withdraw_message to change to our new settings in the eu cookie compliance module settings but the text is applied here in this config file. In the module it uses restricted_html which we don't have anymore, not sure if this is the reason or not, but this can be looked at a later stage.


DuttonMa (approved):
Good stuff, makes sense to me

PR #464, opened 2021-04-23
D8UN-482

DuttonMa:
Changed file upload instructions in README as they were unclear


ooneill547 (approved):
That's much clearer, thanks

PR #474, opened 2021-04-26
D8UN-488 Fiscal Comm: Update custom breadcrumb module to include commissioners

ooneill547:


web/sites/fiscalcommissionni/modules/custom/fiscalcommissionni_breadcrumbs/src/BoardBreadcrumb.php
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build(RouteMatchInterface $route_match) {
+    $breadcrumb = new Breadcrumb();
+    $title_resolver = $this->titleResolver->getTitle($this->request->getCurrentRequest(), $route_match->getRouteObject());
+    if ($this->node) {
+      $links[] = Link::createFromRoute(t('Home'), '<front>');
+      $links[] = Link::fromTextandUrl(t('Commissioners'), Url::fromUri('entity:node/8'));
```
DuttonMa, line 109: We really try to avoid hardcoding node ids like this. Would it be better to search for a basic page node with the title of 'Commissioners' maybe ?

DuttonMa (changes requested):

DuttonMa (approved):
Great stuff, that's much better now

PR #484, opened 2021-04-27
D8UN-511 Fiscal Comm: Add Evidence content type

ooneill547:
- I have more or less duplicated the publications content type to evidence, including adding search api, facets, views etc
- I also had to edit the unity_search_pages module to include the evidence search page.
- Publications content type may be needed in the future, but aren't currently, so I have removed the permissions for authors and supervisors to add content to this content type for now.


soda736 (approved):
LG

DuttonMa (approved):
Looks good, and sounds like a good plan

PR #499, opened 2021-05-06
D8UN-452

soda736:
Updating media wysiwyg migration plugin


web/modules/custom/unity_file_migrations/src/Plugin/migrate/process/MediaWysiwygFilter.php
```diff
@@ -41,39 +44,221 @@
...
+
       try {
+        // Extract the D7 embedded media data.
         $tag_info = $decoder->decode($matches['tag_info'], JsonEncoder::FORMAT);
-        // Convert any 'default' view modes to 'embed'.
-        $tag_info['view_mode'] = str_replace('default', 'embed', $tag_info['view_mode']);
-        return sprintf($replacement_template, $tag_info['view_mode'], $tag_info['fid']);
+
+        // Ensure we have a managed file for the embedded asset.
+        $query = $this->connection->select('file_managed', 'f');
```
omahm, line 101: I think this should be querying against the D7 database but the db object that is injected will be the Drupal 8 connection

omahm (changes requested):
It looks like you might be querying the wrong db.

soda736:
Had a chat with @omahm and decided this code was good to proceed with.

omahm (approved):

PR #504, opened 2021-05-10
D8UN-392

DuttonMa:
Rebuild solr indices completely on each deployment to avoid count errors after a production release


omahm (approved):
Looks good

PR #508, opened 2021-05-13
D8 un 326

soda736:
Removing / from error pages as this directs unauthorised pages to the homepage unstyled.
Amending permissions on Uregni and Odscni for media.


omahm (approved):
LG

PR #511, opened 2021-05-13
D8UN-521 Add upload restrictions to media on unity sites

ooneill547:
ODSCNI and fiscal commission still had the video media type which I removed - this may need to be removed from the install profile if it won't be used in any unity sites


omahm (approved):
Looks good

PR #512, opened 2021-05-13
D8UN-179 Update Cookie banner on all sites

ooneill547:
There was a small spelling error on Uregni and Fiscal Comm, also in the install profile. ODSC wasn't up to date.


omahm (approved):
LG

PR #516, opened 2021-05-17
D8 un 523

DuttonMa:
Move changes to admin/content view and associated tabs over from NIDirect, move quick transitions into 'Operations' column.


omahm (approved):
LG

ooneill547 (approved):

PR #522, opened 2021-05-19
D8UN-529

soda736:
Adding the word Service to all of the cookie compliance banners


ooneill547 (approved):
This should also be updated in the install profile

PR #524, opened 2021-05-19
D8UN-479 Update attachment field to media entity

ooneill547:
Both publication and consultation content types in the install profile had a attachment field that was a file entity, this has been changed to a media entity


soda736 (commented):
LG

soda736 (approved):

PR #528, opened 2021-05-24
D8 un 541

soda736:
We had image and responsive image styles in both the unity install profile and the unity_image_styles module. I have removed all the custom image styles we have in unity_image_styles from the install profile.


ooneill547 (commented):
Just looking at the config files in the unity_image_styles module, I'm not sure if they should have a uuid?

soda736:
I'm not entirely sure, looking through the install profile, some files have uuids and some don't?

DuttonMa:
I'm afraid that I do think that we need to get rid of all of these uuids in config files

DuttonMa (changes requested):
Please remove uuids

soda736:
I've removed all UUIDs from the unity_image_styles module and the unity install profile.

ooneill547 (approved):

DuttonMa (approved):

PR #530, opened 2021-05-24
D8 un 435

soda736:
Issue https://www.drupal.org/project/eu_cookie_compliance/issues/3189864 in the EU cookie compliance module has been fixed in the dev branch which is stable. I have updated unity to use this branch.
There is code to check to see which filter format to use for the popup text fields. We are using basic_html which isn't in the module logic but it then reverts to the filter_fallback_format which in unitys case was plain_text. I have updated the filter settings to use basic_html as the fallback format.


omahm (approved):
Interesting bug. looks good

DuttonMa (approved):

PR #534, opened 2021-05-27
D8 un 525 favicons

soda736:
Adding favicons for Uregni and Fiscal and amending favicon description file for ODSCNI for when we have a favicon.svg.


DuttonMa (approved):
Good work !

PR #553, opened 2021-06-09
D8UN-528 Configure the metatag info on all sites

ooneill547:


omahm (changes requested):
LG but I'd remove user based metatags

ooneill547:
> LG but I'd remove user based metatags

I have restored the user metatags to the default config

omahm (approved):
👍

PR #567, opened 2021-06-14
Changes to unity profile to allow installation to complete

ooneill547:


soda736 (approved):
Sorry, looks like I left those HEAD's in there! 
Looks good.

PR #568, opened 2021-06-14
Update initial config dump for fiscal council

ooneill547:
With changes made to the install profile, the initial config dump for fiscal council needs to be updated


soda736 (approved):
Quick scan through looks good

PR #570, opened 2021-06-15
D8 un 534

soda736:
Removing div class from allowed html content in the basic html text format as it adds unwanted html on a copy and paste.


ooneill547 (approved):
LG

DuttonMa (approved):

PR #571, opened 2021-06-15
D8UN-407 Configure basic page and update permissions

ooneill547:
There were a couple of modules missing that had to be installed also


config/nifiscalcouncil/config/core.entity_form_display.node.basic_page.default.yml
```diff
@@ -17,73 +24,102 @@ mode: default
...
+    third_party_settings: {  }
+    type: metatag_firehose
+    region: content
+  field_next_audit_due:
+    weight: 13
+    settings: {  }
+    third_party_settings: {  }
+    type: datetime_default
+    region: content
+  field_toc_enable:
```
johangant, line 54: Have a chat with @omahm to make sure this is in line with how toc works these days?

omahm: The origins toc update hook will remove the field_toc_enable, Once that update hook has been run you will need to do a config export.

ooneill547: I have updated origins_modules so this has now been removed

johangant (approved):

PR #573, opened 2021-06-16
D8UN-594

soda736:
Fixing regex so it on matches /files/ path. The previous regex was also matching /node/ paths and adding an empty file icon beside it.
It was also stripping out the li tag from the HTML so as we are now being more specific in our regex we don't need the opening <[a-z]{1,3}>

### DIFF
--- a/web/sites/uregni/themes/uregni_theme/uregni_theme.theme
+++ b/web/sites/uregni/themes/uregni_theme/uregni_theme.theme
@@ -207,7 +207,7 @@ function uregni_theme_preprocess_field(&$variables) {
         $body_text = &$variables['items'][0]['content']['#text'];
         // Set a regex pattern to look for body links. We don't want to target body links nested in text.
         // A pattern we want to match is:
-        // <a href="/files/uregni/media-files/24%20May%20board%20minutes%20%28signed%29.pdf"></a>
+        // <a href="/files/uregni/media-files/24%20May%20board%20minutes%20%28signed%29.pdf"></a>.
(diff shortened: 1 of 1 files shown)


web/sites/uregni/themes/uregni_theme/uregni_theme.theme
```diff
@@ -208,7 +208,7 @@ function uregni_theme_preprocess_field(&$variables) {
         // Set a regex pattern to look for body links. We don't want to target body links nested in text.
         // A pattern we want to match is:
         // <div><a href="https://www.uregni.gov.uk/sites/uregni/files/media-files/2019-11-22%20Switchgear%20TNPP%20-%20Final%20Decision.pdf"></a>
-        $embed_regex = '/<[a-z]{1,3}><a[\w\s\.]*href="([^"]*?.([a-z]*))"\>(((?!\<\/a\>).)*)\<\/a\>/';
+        $embed_regex = '/<a[\w\s\.]*href="(\/files\/.*?.([a-z]*))"\>(((?!\<\/a\>).)*)\<\/a\>/';
```
johangant, line 211: I tried this regex at https://regexr.com/ with the example pattern above: `<div><a href="https://www.uregni.gov.uk/sites/uregni/files/media-files/2019-11-22%20Switchgear%20TNPP%20-%20Final%20Decision.pdf"></a>`

It didn't match anything, but the previous one did. Does the example text need updating or does the new regex need refining?

soda736: Yes sorry the comment needs amending, it doesn't need the opening <div>

johangant: Nice one, thanks

johangant (approved):
👍

PR #580, opened 2021-06-21
D8 un 559

soda736:
Migration module
Re-enable content cookie blocker module


web/sites/niauditoffice/modules/custom/niauditoffice_migrations/config/install/migrate_plus.migration.upgrade_d7_field.yml
```diff
@@ -0,0 +1,69 @@
+uuid: 15181210-ece8-4dc8-b79b-23e8f1e6d1c6
```
DuttonMa, line 1: I don't think that we should be saving migrations like 'upgrade_d7_field' as this will migrate actual fields over from D7. It is useful to run it once, but after that we'll keep content type and field definitions in config and won't need to run this again, so it isn't worth saving it in niauditoffice_migrations

web/sites/niauditoffice/modules/custom/niauditoffice_migrations/config/install/migrate_plus.migration.upgrade_d7_field_formatter_settings.yml
```diff
@@ -0,0 +1,144 @@
+uuid: f7478aff-cab3-4dca-b8ab-d2ce4e469094
```
DuttonMa, line 1: Don't need this, same comment as above.

web/sites/niauditoffice/modules/custom/niauditoffice_migrations/config/install/migrate_plus.migration.upgrade_d7_field_instance.yml
```diff
@@ -0,0 +1,113 @@
+uuid: 25a649d6-2ebb-43be-a194-1d2d24280ecd
```
DuttonMa, line 1: Don't need this, same comment as above.

web/sites/niauditoffice/modules/custom/niauditoffice_migrations/config/install/migrate_plus.migration.upgrade_d7_field_instance_widget_settings.yml
```diff
@@ -0,0 +1,123 @@
+uuid: d7892222-3a78-417f-8684-d1d3c45e9c61
```
DuttonMa, line 1: Don't need this, same comment as above.

web/sites/niauditoffice/modules/custom/niauditoffice_migrations/config/install/migrate_plus.migration.upgrade_d7_filter_format.yml
```diff
@@ -0,0 +1,61 @@
+uuid: bd898e10-926f-4778-ae92-8fea88d13bbf
```
DuttonMa, line 1: Don't think we need this one either.

web/sites/niauditoffice/modules/custom/niauditoffice_migrations/config/install/migrate_plus.migration.upgrade_d7_filter_settings.yml
```diff
@@ -0,0 +1,29 @@
+uuid: dee16918-f4b2-4109-a417-32dda9451339
```
DuttonMa, line 1: Don't need this, same comment as above.

web/sites/niauditoffice/modules/custom/niauditoffice_migrations/config/install/migrate_plus.migration.upgrade_d7_image_styles.yml
```diff
@@ -0,0 +1,37 @@
+uuid: e512bcea-f006-45ee-8952-2a6876afe19c
```
DuttonMa, line 1: Don't need this, same comment as above.

DuttonMa (changes requested):
Have picked out a few files but there are plenty more that should come out too. The basic idea is that any migrations that are to do with 'structure' are run once only and not kept. We only keep migrations that we might want to run several times to migrate data that might changes e.g. nodes, taxonomy terms. 
Have a look at Uregni as an example and see what migrations we have saved in the uregni_migrations module.

soda736:
I've ran the migrations and have a database so I have now deleted the migration files we no longer require.

DuttonMa:
Looks better but don't think you're quite there yet. As a general rule of thumb, anything that has a name with 'field instance' in it is not needed.

soda736:
I've removed the field instance metatag files so I think that's all the field_instance files deleted.

DuttonMa (approved):
All looks good to me now

PR #582, opened 2021-06-21
D8 un 548

DuttonMa:
Changes to ensure that Google Analytics keys are not lost from Unity sites during a release.
Google Analytics keys should be stored in Platform.sh environment variables in the format:

subsite_id_GOOGLE_ANALYTICS_KEY 
(e.g. UREGNI_GOOGLE_ANALYTICS_KEY)

If these environment variables are set up correctly then they will be used to set the Google Analytics key for the site after each config import.

In order for these changes to work fully (and for the hook_config_ignore_settings_alter to fire) the config_ignore module must be installed on all Unity sites, changes have been included to ensure that.


soda736 (approved):
LGTM but maybe better with Steve or Johan running their eye over this.

omahm (approved):
Nice one, good job 👍

PR #591, opened 2021-06-23
Setting solr rebuild to not timeout until 30 mins

soda736:


DuttonMa:
Perhaps it would be a good idea to comment out the 'sapi-rt' commands in this PR as well ?

soda736:
Yes, good idea, I have commented out the sapi-rt command and the solr rebuild has completed.

DuttonMa (approved):
Looks good to me

PR #600, opened 2021-06-28
Circleci issues

soda736:
Add latest code to all sites.
I think this could be tidier.
Ideally I would like to have $site uregni odscni etc line 311
then cd web/sites/$site/themes/$site_theme line 314
But circleci doesn't like this concat.
I've tried adding in parameters and calling them:
parameters
 type: string
 default: "_theme"
cd web/sites/$site/themes/$site<<parameters.theme>>
But it didn't like this either.


DuttonMa (changes requested):
As discussed, I think that this would be a better option:

cd web/sites/$site/themes
cd *theme

soda736:
Yes, I've tested this and it works great! Thanks

DuttonMa (approved):

PR #606, opened 2021-06-29
Niao permissions

soda736:
Updating the modules and permissions needed for NIAO


config/niauditoffice/config/google_analytics.settings.yml
```diff
@@ -0,0 +1,37 @@
+account: ''
```
ooneill547, line 1: This needs to be changed to `account: <placeholder>`

ooneill547 (changes requested):
Just that small change in google analytics

config/niauditoffice/config/google_analytics.settings.yml
soda736: Good spot, thanks.

ooneill547 (approved):

PR #613, opened 2021-06-30
D8 un 588

soda736:
Search page templates, preprocessing and regions.

### DIFF
--- a/web/sites/niauditoffice/themes/niauditoffice_theme/niauditoffice_theme.theme
+++ b/web/sites/niauditoffice/themes/niauditoffice_theme/niauditoffice_theme.theme
@@ -15,11 +15,11 @@ function niauditoffice_theme_theme_suggestions_views_view_alter(array &$suggesti
 }
 
 /**
-* Implements hook_theme_suggestions_HOOK_alter().
-*/
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
 function niauditoffice_theme_theme_suggestions_field_alter(array &$suggestions, array $variables) {
   if (isset($variables['element']['#view_mode'])) {
-  $suggestions[] = 'field__node__' . $variables['element']['#field_name'] . '__' . $variables['element']['#view_mode'];
+    $suggestions[] = 'field__node__' . $variables['element']['#field_name'] . '__' . $variables['element']['#view_mode'];
   }
 }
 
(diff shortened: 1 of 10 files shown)


web/sites/niauditoffice/themes/niauditoffice_theme/niauditoffice_theme.theme
```diff
@@ -0,0 +1,86 @@
...
+  // A list of search page titles on the Uregni site.
+  $search_pages = [
+    'News',
+    'Publications',
+    'Search',
+  ];
+
+  // If a search page set the variable to true.
+  foreach ($search_pages as $search_page) {
+    if ($variables['page']['#title'] == $search_page) {
```
omahm, line 41: This works but take a look at https://www.php.net/manual/en/function.in-array.php
Could replace 5 lines with 1.

ooneill547 (approved):

web/sites/niauditoffice/themes/niauditoffice_theme/templates/views/views-view-unformatted--search-result-page.html.twig
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @see template_preprocess_views_view_unformatted()
+ */
+#}
+{% if title %}
+  <h3>{{ title }}</h3>
+{% endif %}
+{% for row in rows %}
+  {%
+    set row_classes = [
```
omahm, line 23: row_classes will be set with every look of the rows, I'd create this outside the loop

omahm (changes requested):
Looks good overall. A few wee suggestions..  ping me if you want to go over in_array

web/sites/niauditoffice/themes/niauditoffice_theme/templates/views/views-view-unformatted--search-result-page.html.twig
soda736: I am trying to add the search result class to every article in the view. There is code for this in the other unity sites and styling in place for the .search-results. Would it be better to have the search result class outside of the rows and add the styling to .search-results article?

web/sites/niauditoffice/themes/niauditoffice_theme/niauditoffice_theme.theme
soda736: Think I know what you mean, I've made a change to it.

omahm (approved):
Yup, that's it. Good stuff

PR #630, opened 2021-07-07
D8UN-409 Fiscal Council: Theme front page

ooneill547:


config/nifiscalcouncil/config/core.extension.yml
```diff
@@ -26,6 +26,7 @@ module:
   ctools: 0
   datetime: 0
   datetime_range: 0
+  devel_generate: 0
```
soda736, line 29: Maybe uninstall this module?

soda736 (approved):
Just one small comment to maybe uninstall devel generate, otherwise looks good.

config/nifiscalcouncil/config/core.extension.yml
ooneill547: Good catch - I'll do that now

PR #652, opened 2021-07-19
D8 un 580

soda736:


omahm (approved):
Looks good. A quicker way to determine the node type is to call bundle() on the node object.

soda736:
No worries, cheers, I've changed it to bundle().

PR #664, opened 2021-07-22
D8UN-606 fix image mappings

johangant:


omahm (approved):

ooneill547 (approved):
Great work! It seems to be working on uregni, we just need to apply the same changes to the other sites

johangant:
> Great work! It seems to be working on uregni, we just need to apply the same changes to the other sites

I'll get this rolled out to the other active sites as part of this work. Cheers!

PR #665, opened 2021-07-22
D8UN-614 Fix workbench area views on all sites

ooneill547:
I needed to add the next audit due field to the audit office site and configure it so that the view could be configured properly


omahm (approved):
LG

PR #666, opened 2021-07-23
D8UN-634 ~ Update Core to 9.2.2, Update file transliteration patch

omahm:
- Core update to 9.2.2. 
- Transliteration patch will need to be thoroughly tested. I carried out a test using a file called 'test file @test.txt' and it renamed it to 'test-file-test.txt.'


omahm:
Update in a later PR

PR #669, opened 2021-07-23
D8UN-623 Add current page to all breadcrumb links

ooneill547:
This needed to be completed for all unity sites


web/sites/fiscalcommissionni/modules/custom/fiscalcommissionni_breadcrumbs/src/PageBreadcrumb.php
```diff
@@ -0,0 +1,114 @@
...
+    return $match;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build(RouteMatchInterface $route_match) {
+    $breadcrumb = new Breadcrumb();
+    $title_resolver = $this->titleResolver->getTitle($this->request->getCurrentRequest(), $route_match->getRouteObject());
+    if ($this->node) {
```
omahm, line 105: Shouldn't really need this check as build will only get called if we have passed the applies() call

omahm (changes requested):
Generally looks good, just a couple of issues.

ooneill547:
> Generally looks good, just a couple of issues.

I have updated the fiscal commission breadcrumbs to check if it's correct and then I will do the rest

johangant:
I think this could do with an update/merge from `development` to assess what has changed; there are a number of things that look like they come from other feature branches and should already be in `development` by now.

omahm (approved):
Looks good

PR #670, opened 2021-07-26
D8UN-617 CKEditor embedded entity styles

johangant:
* Moving ckeditor SCSS from Urengi to Unity theme (https://github.com/dof-dss/nicsdru_unity_theme/pull/314)
* Add Lando command for npm usage
* Drop two image styles from a thwoop view mode that look like they belong to a similar, non-thwoopy, view mode.


ooneill547 (approved):

johangant:
I've removed these from uregni, but I notice that sites like nifiscalcouncil have overrides and per-site adjustments which would not be safe to remove without a more detailed understanding of their designs and why those overrides where needed.

Proposing to merge as-is and other existing sites should either override Unity's styles + future sites can add in only the small overrides they need, if any.

PR #681, opened 2021-07-30
D8UN-628 HTML document feature

johangant:
Relates to: https://github.com/dof-dss/nicsdru_unity_theme/pull/319

~NB: currently only config in place for Uregni. I expect this to expand to other sites as needed once refined/reviewed.~

* NI Audit Office site: new config; body field replaced with dedicated HTML field. Makes it easier to monitor table bloat and also means we can avoid the summary option which the body field presently requires.
* Added a custom icon for CKEditor embed button, to replace the default 'E' icon for entity embed.
* TODO: 
** DI for whatlinkshere service
** a bit more refinement/checks in the Unity HTML pubs module.
** Refinement in the breadcrumb builder class.
** Config for Full HTML text format.


omahm (approved):
Looks good

PR #683, opened 2021-08-02
D8UN-643 OCTF: Set up news content type

ooneill547:


config/octf/config/search_api.index.news.yml
```diff
@@ -0,0 +1,117 @@
+uuid: bb8cbe60-082e-47e2-b048-f5186eb76318
```
johangant, line 1: Interesting. I could well be wrong or missing other context, but is there a general content index for the site? An index per content type would usually either be a sign of some specific requirements difference (eg: NIDirect's health conditions) or a possible anti-pattern where news could easily be bundled with other content into the same index and filtered through the view config.

YMMV but thought I'd check, just in case :)

ooneill547: There will be a general content index for the site, but I haven't created it yet. This is the way we have created all of the searches on the unity sites. The news page will have facets and an exposed search field. 

Do you think this would be better done on all sites as one general search index?

johangant: > Do you think this would be better done on all sites as one general search index?

This is how I'd start as a default. Generally speaking, separate indexes are useful if you need to vary how they behave (ie: their schema differs) but you can do a lot with query preprocess commands alone. Most sites I see tend to be single Solr service + single index, with entities of varying content types added to it. 

Facets would be one to check but from what you're describing, at a high level, separate indexes sounds like it might be adding config where you don't need it. You might in future, but I'd split it out and add the extra complexity when you need it rather than in advance.

ooneill547: I'll try and reconfigure how this site works and hopefully it will work with this approach. It would be a lot less work in the long run for future sites.

DuttonMa (approved):

PR #686, opened 2021-08-05
CI refinements and improvements

johangant:
- Streamline the edge build workflow
- Put PSH CLI tool into $PATH, adjust post-deploy job steps
- Refine static analysis jobs
- Remove redundant kernel/unit tests
- Remove commented out steps
- Drop unwanted config


.circleci/config.yml
```diff
@@ -41,6 +37,23 @@ commands:
           name: Fetch dependencies with composer
           command: |
             composer install --no-interaction --optimize-autoloader
+  composer_tasks__edge_packages:
```
johangant, line 40: Moved this task into its own command to keep it DRY. Previously, and on NIDirect, this kind of operation was run inside individual jobs resulting in duplication + general problems with entropy

.circleci/config.yml
```diff
@@ -41,6 +37,23 @@ commands:
...
+                          dof-dss/nicsdru_origins_profile:dev-development
+  install_psh_cli:
+    description: "Install the Platform.sh CLI tool"
+    steps:
+      - run:
+          name: Install the Platform.sh CLI tool
+          command: curl -sS https://platform.sh/cli/installer | php
+      - run:
+          name: Add platform cli tool to $PATH
+          command: echo 'export PATH="$HOME/"'.platformsh/bin':"$PATH"' >> $BASH_ENV
```
johangant, line 56: This step ensures the tool can be called without using its absolute system path, which makes things a lot more readable.

This step to `>> $BASH_ENV` is a Circle CI concern and by passing the output back out into the shell environment it means that `$PATH` variable is accessible to later job steps.

.circleci/config.yml
```diff
@@ -51,13 +64,16 @@ commands:
             sudo cp /home/circleci/project/.circleci/docker-php-circleci.ini /usr/local/etc/php/conf.d/
             sudo apt-get update
             sudo apt-get install -y libpng-dev
-            sudo docker-php-ext-install gd
-  github_hosts_workaround:
+            sudo docker-php-ext-install gd pcntl posix
```
johangant, line 67: Packages needed by PSH CLI tool

.circleci/config.yml
```diff
@@ -51,13 +64,16 @@ commands:
             sudo cp /home/circleci/project/.circleci/docker-php-circleci.ini /usr/local/etc/php/conf.d/
             sudo apt-get update
             sudo apt-get install -y libpng-dev
-            sudo docker-php-ext-install gd
-  github_hosts_workaround:
```
johangant, line 55: Renamed this as it's not just GitHub, nor is it a workaround.

.circleci/config.yml
```diff
@@ -72,22 +88,33 @@ jobs:
...
       - image: circleci/php:7.4.15-apache-browsers
+    environment:
+      PROJECT_ROOT: "/home/circleci/project"
     steps:
       - attach_workspace:
           at: ./
       - run: sudo cp /home/circleci/project/.circleci/docker-php-circleci.ini /usr/local/etc/php/conf.d/
       - run:
           name: PHPCS analysis
-          command: /home/circleci/project/phpcs.sh /home/circleci/project "/home/circleci/project/web/sites /home/circleci/project/web/modules/origins /home/circleci/project/web/modules/custom /home/circleci/project/web/themes/custom /home/circleci/project/web/profiles/unity /home/circleci/project/web/profiles/origins"
```
johangant, line 85: Much more readable and maintainable now with one path per line, then fed into the command parameters as a shell variable.

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
           name: Deprecated code check
           command: |
             cd /home/circleci/project
-            vendor/bin/drupal-check /home/circleci/project/web/modules/custom /home/circleci/project/web/modules/origins
-
-  # Run any unit tests and any kernel tests against a vanilla D8 site + our site config imported over it (no predefined content).
-  unit_kernel_tests:
```
johangant, line 102: Redundant. All of this is now covered, at a functional level, in the process to build the edge environment.

johangant: Which is just as well because, as the author of this stuff, I can say it's pretty ugly even though it might just about be functional.

In future, it's probably more valuable to run builds directly on platform.sh as that gives us the same containers, config and data... whereas Circle CI costs us a bit each time and is only an approximation (they're very generic containers)

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
-          command: |
-            cd /home/circleci/project/web
-            # for site in uregni liofa fictcommission
-            for site in uregni odscni fiscalcommissionni niauditoffice
-            do
-              ../vendor/bin/drush -l $site st | grep "Drupal bootstrap : Successful"
-            done
   # Nightly edge build
-  unity_edge_build:
+  edge_build:
```
johangant, line 131: Removed project name from the job name - felt surplus to requirement as the project name is in, well, the repository name :)

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
     docker:
-      - image: circleci/php:7.4.15-apache-browsers
+      - image: circleci/php:7.4.15-apache-node-browsers
```
johangant, line 133: Swapped image so we get node.js and npm installed out of the box; so fewer extra install steps needed after.

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
     environment:
       # git variables to avoid empty committer identity errors
       EMAIL: "circleci@localhost"
       GIT_COMMITTER_NAME: "Circle CI"
       GIT_AUTHOR_NAME: "Circle CI"
+      EDGE_BUILD_BRANCH: "D8UN-edge"
```
johangant, line 139: No reason to be coy about this by hiding it in a Circle CI env var. It's pretty easy to spot in the public repo anyway!

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
-          name: Add extra OS and PHP extensions/config
-          command: |
-            sudo docker-php-ext-install pcntl posix
-      - run:
-          name: Keyscan Platform.sh region hostnames + GitHub for password-less access
-          command: |
-            mkdir -p ~/.ssh
-            ssh-keyscan -H ssh.$PLATFORM_REGION >> ~/.ssh/known_hosts
-            ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+          name: Switch to edge branch
```
johangant, line 147: There are two ways to do this...

1. Clone on development, checkout edge, update composer packages + push to edge.
2. Just clone edge, update composer and push back to edge.

The problem with 2 is that you either have to merge in development or you potentially miss other package updates which would eventually make for a build that's out of kilter with the rest of things.

So option 1 is the safest, even if it might look a bit more verbose.

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
+          command: git checkout $EDGE_BUILD_BRANCH
+      - composer_tasks__edge_packages
       - run:
-          name: Install the Platform.sh CLI tool
+          name: Push composer package updates back to GitHub
           command: |
-            curl -sS https://platform.sh/cli/installer | php
-      - run:
-          name: Re-point dof-dss packages to use latest development code and push.
```
johangant, line 299: Handled earlier via a command rather than inline run statements.

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
             git commit -m "Set dof-dss packages to HEAD development for build"
-
+      - run:
+          name: Re-build site theme npm dependencies
+          command: |
             # Need to re-build site theme with any Unity changes.
-            # for site in uregni liofa fictcommission
-            for site in uregni odscni fiscalcommissionni niauditoffice
+            # Download and configure node and npm.
+            sudo apt install libjpeg-dev automake
```
johangant, line 160: Moved out of the previous loop because it's a system concern not a theme folder concern.

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
+            for site in `ls -l ~/project/web/sites | grep ^d | awk '{print $9}'`
```
johangant, line 161: This like lets us scan the sites folder for directory names rather than list them by hand. So adding a site in future shouldn't need any more maintenance here.

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
-
-              # Build site themes.
-              npm install
-              npm rebuild node-sass
-              npm run build
-
-              git add *.css
-              git commit -m "Theme rebuild"
-              cd ~/project
+              if [[ -d ~/project/web/sites/${site}/themes/${site}_theme ]]; then
```
johangant, line 163: Some sites in the repo exist, but are shells and have no theme folder. This checks whether there is something like `uregni_theme` before proceeding to do theme 'stuff'

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
+                cd ~/project/web/sites/${site}/themes/${site}_theme
+                sed -i -e "s/\"nicsdru_unity_theme.*/\"nicsdru_unity_theme\": \"github:dof-dss\/nicsdru_unity_theme#development\",/" package.json
```
johangant, line 165: Each theme, currently, needs its own dependencies repointed. I know this is under review right now so expect this to change when that piece of work is ready!

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
+                npm install
+                npm rebuild node-sass
+                npm run build
+              else
+                echo "No theme folder detected under ~/project/web/sites/${site}/themes/${site}_theme, skipping."
```
johangant, line 170: Debug statement to minimise WTF moments in future

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
             done
 
             # Push to fixed, non-integrating build branch. GitHub webhook integration will propagate this
             # to platform.sh for later steps to use.
             git push -f origin $EDGE_BUILD_BRANCH
 
-            # Pause for webhook to propagate the change from GH to PSH.
-            sleep 10s
-      - run:
-          name: Activate the environment if not found.
```
johangant, line 341: Seems surplus to requirement now. If it's not active the very first time, then someone can just go an activate it by hand. Literally a one time task so see no need to check for it here even if we can.

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
-            /home/circleci/.platformsh/bin/platform drush 'pmu fastly -l liofa'  -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
-            /home/circleci/.platformsh/bin/platform drush 'pmu fastly -l fictcommission'  -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
-            /home/circleci/.platformsh/bin/platform drush 'pmu fastly -l employmenttribunalsni'  -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
-            /home/circleci/.platformsh/bin/platform drush 'pmu fastly -l attorneygeneralni'  -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
-            /home/circleci/.platformsh/bin/platform drush 'pmu fastly -l fiscalcommissionni'  -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
-            /home/circleci/.platformsh/bin/platform drush 'pmu fastly -l niauditoffice'  -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
+            for site in `ls -l ~/project/web/sites | grep ^d | awk '{print $9}'`
+            do
+              result=$(platform drush "st -l ${site}" -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH | (grep "Drupal bootstrap : Successful" || true) )
```
johangant, line 203: Circle CI uses pipefail options which means if a command in a chain/pipe fails then the entire job exits.

Grep has some unusual return codes when it matches/doesn't match patterns which threw this check off a few times. By wrapping it up with `|| true` we guard against exit codes that otherwise clash with Circle CI's command evaluation and it means the loop can continue even if we can't get a successful bootstrap from a site that might have a directory.

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
-            /home/circleci/.platformsh/bin/platform drush 'cim -l niauditoffice'  -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
+            for site in `ls -l ~/project/web/sites | grep ^d | awk '{print $9}'`
+            do
+              result=$(platform drush "st -l ${site}" -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH | (grep "Drupal bootstrap : Successful" || true) )
+              if [[ ! -n $result ]]; then
+                platform drush "cim -l ${site}" -y -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH
+              fi
+            done
       - run:
           name: Force purge of Solr index
```
johangant, line 229: Unclear which sites need this done right now. Taking it out so it can be done manually, as needed, until we get more clarity

.circleci/config.yml
```diff
@@ -96,392 +123,124 @@ jobs:
...
-      - run:
-          name: Install the Platform.sh CLI tool
-          command: |
-            curl -sS https://platform.sh/cli/installer | php
-      - run:
-          name: Force purge of Solr index
-          command: /home/circleci/.platformsh/bin/platform ssh -p $PLATFORM_PROJECT_ID -e $EDGE_BUILD_BRANCH "curl http://${PLATFORM_SOLR_HOST_UREGNI}/solr/uregni_index/update --data '<delete><query>*:*</query></delete>' -H 'Content-type:text/xml; charset=utf-8' && curl http://${PLATFORM_SOLR_HOST_UREGNI}/solr/uregni_index/update --data '<commit/>' -H 'Content-type:text/xml; charset=utf-8'"
+            for site in uregni
+            do
+              uc_site=$(echo ${site} | tr '[:lower:]' '[:upper:]')
```
johangant, line 233: There's some faff here to take the site name and uppercase it with the `tr` command before its fed into the curl command.

.circleci/config.yml
```diff
@@ -531,27 +265,17 @@ workflows:
...
+  nightly-edge-build-post-build-tasks:
     triggers:
       - schedule:
           # At 01:23 Monday to Friday
           cron: "23 1 * * 1-5"
           filters:
             branches:
               only:
-                - development
+                - D8UN-edge
```
johangant, line 279: Switched because we perform these tasks on the edge build using code from the edge branch.

.circleci/config.yml
ooneill547: I think this was implemented as we were getting inflated results on the search pages on uregni. As the site is now live it's probably not a big issue.

ooneill547 (approved):
Your comments and changes seem to make sense, to my limited understanding

PR #698, opened 2021-08-13
D8UN-653 Fiscal Council: Add sidemenu

ooneill547:


web/sites/nifiscalcouncil/themes/nifiscalcouncil_theme/nifiscalcouncil_theme.theme
```diff
@@ -79,3 +80,45 @@ function nifiscalcouncil_theme_preprocess_form(array &$variables) {
...
+    $search_pages = [
+      'news',
+      'publications',
+    ];
+
+    $plugin_id = $variables['plugin_id'];
+
+    foreach ($search_pages as $search_page) {
+      // Search the machine name for search page name and set variable to this.
+      if (str_contains($plugin_id, $search_page)) {
```
omahm, line 108: This is a PHP 8 method and you won't be able to use this on the site (7.4). I'd suggest using strpos.

omahm (changes requested):
Mostly ok but just need to replace the str_contains call.

web/sites/nifiscalcouncil/themes/nifiscalcouncil_theme/nifiscalcouncil_theme.theme
ooneill547: I removed this section of code because it wasn't actually being used in this site.

omahm (approved):
LG

PR #711, opened 2021-08-18
D8UN-672 USI: Configure news content type

ooneill547:
I had to delete the photo field from other content types to recreate it as a media entity


soda736 (approved):
LG

PR #720, opened 2021-08-20
Add production config split for NIAuditOffice site

johangant:


config/niauditoffice/production/fastly.settings.yml
```diff
@@ -0,0 +1 @@
+site_id: ''
```
DuttonMa, line 1: Think we need to set a site id here really

DuttonMa (changes requested):

config/niauditoffice/production/fastly.settings.yml
johangant: @David-Corbett @ooneill547 is this something you can set up/add in fast.ly, please?

ooneill547: I don't have access to fastly to get things like this, sorry

johangant: Okay, will flag separately. I tried looking for config settings using drush but it's not presently enabled on the site so there's no existing active config to use here. It'll need someone with fastly config access to set up first before we can advance this PR

PR #726, opened 2021-08-24
D8 un 680

DuttonMa:
Added code to derive page title from route after $this->request->get('_title') stopped working for an unknown reason !

I'm not completely happy with this solution tbh, but we need to get a fix in fairly quickly as it is holding up the Unity release. I will continue to look for a better solution.


ooneill547 (approved):

web/modules/custom/unity_search_pages/src/Controller/SearchPagesController.php
```diff
@@ -86,4 +86,34 @@ public function getTitle($route = NULL) {
...
+   */
+  public function getTitleFromRoute($route = NULL) {
+    // For example, route
+    // view.publications_search.publication_search_page
+    // gives the title 'Publications'.
+    $title = "";
+    $route_parts = explode('.', $route);
+    if ((count($route_parts) > 2) && isset($route_parts[2])) {
+      $title = $route_parts[2];
+      $title = str_replace('search_page','',$title);
```
omahm, line 103: str replace can use an array of values, so you could replace this with:
$title = str_replace(['search_page', '_'],'',$title);

PR #750, opened 2021-09-09
D8UN-643 OCTF: Style news page

ooneill547:


web/sites/octf/themes/octf_theme/octf_theme.theme
```diff
@@ -17,3 +20,118 @@ function octf_theme_preprocess_views_view(array &$variables) {
...
+      $variables['attributes']['class'][] = 'sidebar-second';
+    }
+  }
+}
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function octf_theme_theme_suggestions_field_alter(array &$suggestions, array $variables) {
+  if (isset($variables['element']['#view_mode'])) {
```
johangant, line 40: Contrib and core code tends to prefer `empty()` over `isset()` as the latter can still return true for a value like 0, because the variable has been set/initialised. Empty provides a more robust check for 'truthiness' or nulls.

See https://www.drupal8.ovh/en/tutoriels/228/isset-vs-empty-vs-null

johangant (changes requested):
A nitpick but I'd recommend switching the `isset()` call to `empty()` for fewer unexpected truthy evaluations in future. Otherwise, great!

web/sites/octf/themes/octf_theme/octf_theme.theme
ooneill547: Thanks, I'll create a ticket to change this on the other sites.

johangant (approved):

PR #759, opened 2021-09-15
Document plugin

soda736:


web/modules/custom/unity_file_migrations/src/PublicationLinkProcessor.php
```diff
@@ -0,0 +1,77 @@
...
+      if (!empty($embedded_file_name)) {
+        // There may be a better way to do this. The filename taken from the body text relates to the uri of the file
+        // some uri's are stored under public:// and some public://media-files. I couldn't find a way to add both or
+        // do a strpos() in loadByProperties() but this works.
+        $files = \Drupal::entityTypeManager()
+          ->getStorage('file')
+          ->loadByProperties(['filename' => urldecode($embedded_file_name)]);
+
+        if (empty($files)) {
+          $files = \Drupal::entityTypeManager()
```
johangant, line 49: Not a deal breaker, but usually services are passed into classes through dependency injection which lets the service container add the correct class based on wider app context.

The end result is you see something like `$this->serviceVariable->method()` rather than `\Drupal::service('servicename')->method()`

Both work, but the former is more maintainable and testable.

web/modules/custom/unity_file_migrations/src/PublicationLinkProcessor.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $file = reset($files) ?: NULL;
+        // Ensure we have a matched file to work with and store the file ID in an array.
+        if (isset($file)) {
+          $file_target_id = $file->id();
+          array_push($file_target_ids, $file_target_id);
+        }
+        // This is handy for debugging if the filename doesn't match.
+        else {
+          var_dump('Nid: ' . $nid . ' filename: ' . $embedded_file_name);
```
johangant, line 67: I think there's a service in migrate, or if not definitely in core, which lets you pass migration messages into. That's handy because it ties up with wider migrate logging and output settings for Drupal as a whole.

johangant (approved):
I've approved, but there are some non-critical things to consider when writing the classes. It's a good habit to work with Drupal services and understand how to add them to custom classes, even if the static calls can also do the immediate task at hand.

web/modules/custom/unity_file_migrations/src/PublicationLinkProcessor.php
johangant: Hadn't seen that this was a static function. DI won't be necessary here. My fault, sorry!

PR #764, opened 2021-09-16
D8UN-624

soda736:
Amending the frontpage view custom cache tag.

### DIFF
--- a/config/niauditoffice/config/views.view.frontpage.yml
+++ b/config/niauditoffice/config/views.view.frontpage.yml
@@ -8,6 +8,7 @@ dependencies:
   module:
     - node
     - user
+    - views_custom_cache_tag
 _core:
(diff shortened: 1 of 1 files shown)


config/niauditoffice/config/views.view.frontpage.yml
```diff
@@ -25,8 +26,9 @@ display:
         options:
           perm: 'access content'
       cache:
-        type: none
-        options: {  }
+        type: custom_tag
+        options:
+          custom_tag: ''
```
ooneill547, line 31: should there not be a custom tag included here?

soda736: No we want to leave this empty as this will then stop the node_list tag from being displayed. It doesn't have a content type associated with it so leaving it empty works.

ooneill547: No problem, I wasn't sure

ooneill547 (approved):

PR #773, opened 2021-09-28
D8UN-706

DuttonMa:
Check that Fastly module is installed before uninstalling


ooneill547 (approved):

johangant (approved):
Perfect - checks for a working Drupal install as well as only uninstalling fastly if it's enabled in the first place. THanks for this

PR #788, opened 2021-10-05
D8UN-724

soda736:
Employment trib migration module


web/sites/employmenttribunalsni/modules/custom/employmenttribunalsni_migrations/config/install/migrate_plus.migration.upgrade_d7_node_settings.yml
```diff
@@ -0,0 +1,29 @@
+uuid: bb1236b1-dac6-4d44-ad1f-0adf23a46130
```
DuttonMa, line 1: Don't think we usually include 'node_settings' as it is config and shouldn't need to be run again

DuttonMa (changes requested):
Just a few files that I don't think should be in here, otherwise looks good

soda736:
I've removed all the mentioned migration files.

PR #790, opened 2021-10-05
D8UN-713 MAH: Configure News content type. 

ooneill547:
Original photo image field removed from all content types so it can be added as an entity reference.


soda736 (approved):
LG

PR #861, opened 2021-11-02
D8UN-601 Solr sites search reconfigure

ooneill547:
Alter NIAO search api index fields to match Uregni and set caching to non for site search on both sites.

### DIFF
--- a/config/uregni/config/views.view.search.yml
+++ b/config/uregni/config/views.view.search.yml
@@ -28,7 +28,7 @@ display:
         type: none
         options: {  }
       cache:
-        type: search_api_tag
+        type: none
         options: {  }
       query:
         type: views_query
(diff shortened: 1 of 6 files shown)


config/uregni/config/views.view.search.yml
```diff
@@ -28,7 +28,7 @@ display:
         type: none
         options: {  }
       cache:
-        type: search_api_tag
+        type: none
```
johangant, line 31: I recall you mentioned this was needed for something last week. As long as this is set intentionally then that's fine... if it's an accident feel free to swap it back :)

johangant (approved):
Looks good @ooneill547 - just flagged the one issue around removal of caching from the views display. I think it's intentional based on last week's discussions

config/uregni/config/views.view.search.yml
ooneill547: Yes, the search highlight retains the original search from the last cache clear with this set to search_api_tag. It gives weird results if you try multiple search terms within 5 minutes. Martin reviewed the other sites with this setting earlier today and seems to think it is fine.

PR #891, opened 2021-11-12
Ire misc

soda736:


omahm (approved):
I was worried when I saw 97 file changes, thankfully most are just deletions!

PR #900, opened 2021-11-17
D8 un 696u

soda736:


ooneill547 (commented):
If this is theming that will be used in all of the unity sites, it might be better to have the stylesheet in unity theme as a partial which is automatically pulled in to all sites, like '_breadcrumb.scss'. Either that or include the library in the unity theme. Either option would mean that we don't need to add the library into every site.

soda736:
I have removed the sub-theme specific libraries for the sidebar-menu styling and the field label as these will be useful on all sites. Adding them into the unity_theme.

ooneill547 (approved):

PR #901, opened 2021-11-17
Config ignore settings for social media links

soda736:
We don't want the social media config to be overwritten on a deployment so adding the settings in to be ignored.


johangant:
No qualms about the way this is excluded from config import. It's unusual to see so many changes being made to social media handles in this way... is this being used correctly?

johangant:
Discussed with @soda736 - this does offer a way to update these values without triggering a deployment, even if it's a pretty infrequent event. No value in rethinking or excluding this now the work has been done.

johangant (approved):

PR #919, opened 2021-11-22
D8 un 805

DuttonMa:
Add new publication type to search index, add a facet and display it on the publications search page.


omahm (approved):
LG

soda736 (approved):
LG

PR #921, opened 2021-11-22
MAHI: Configure footer contact block properly

ooneill547:
This kept failing on platform, so hopefully this will fix it permanently.


omahm (approved):
LG

PR #922, opened 2021-11-25
D8 un 805

DuttonMa:
Migration changes for publication 'type' field (which has changed from a fixed list to a taxonomy ref)


soda736 (approved):
LG

ooneill547 (approved):

PR #927, opened 2021-11-26
D8 un 686 install

soda736:
Updating the install profile to something more like we have in development.


ooneill547 (approved):
There is a lot of files to look through, so I haven't checked it all, but from what I've seen it looks good. Have you tried spinning it up locally to make sure it works with no issues?

soda736:
Yes, I have spun it up, too many times!! :) but yes it works. not sure why but the workflow config file has to be in both install and optional directories for it to work. Alot of the content types have Summary fields which we haven't used up to now but I have left those in incase we ever need them.

ooneill547:
> Yes, I have spun it up, too many times!! :) but yes it works. not sure why but the workflow config file has to be in both install and optional directories for it to work. Alot of the content types have Summary fields which we haven't used up to now but I have left those in incase we ever need them.

Brilliant! Well done, this is a massive piece of work which will make life a good bit easier going forward.

PR #936, opened 2021-12-02
D8UN-809

DuttonMa:
Drupal core 9.2.10 upgrade


ooneill547:
Should we hold of on this for release today?

DuttonMa:
> Should we hold of on this for release today?

Yes, good thought, we'll let it run in the overnight edge build before releasing it

ooneill547 (approved):

composer.lock
```diff
@@ -1906,16 +1906,16 @@
         },
         {
             "name": "dof-dss/nicsdru_unity_theme",
-            "version": "v0.1.39",
+            "version": "v0.1.38",
```
ooneill547, line 1909: your local version of unity theme needs updated, this might cause issues.

DuttonMa: Great catch ! Have fixed that now

PR #954, opened 2021-12-08
D8 un 862

DuttonMa:
Create a bash script to run drush commands on all Unity sites


ooneill547 (approved):
Looks much better than my local attempt

omahm (approved):
Nice one.

PR #960, opened 2021-12-10
D8 un 861

DuttonMa:
Split search api index into separate indices for news and publications - this is better for caching as it means that the news cache is not cleared when you add or edit a publication, for example.


omahm (approved):
LG

PR #961, opened 2021-12-10
Search preprocessing

soda736:
Moving duplicated search page code that occurs across our .theme files to the search module. Some remnants of block preprocessing are removed too as they are in the nicsdru_unity_theme.theme file now.


omahm (approved):
Nice one and good example of DRY

PR #968, opened 2021-12-14
D8UN-861

DuttonMa:
Split Urology search index into separate search indices for publications and news, re-work views and facets accordingly


config/urologyservicesinquiry/config/views.view.news_search.yml
```diff
@@ -27,7 +27,7 @@ display:
         type: none
         options: {  }
       cache:
-        type: search_api_tag
+        type: tag
```
ooneill547, line 30: Should this not be search_api_tag?

DuttonMa: Ah yes, well spotted !

config/urologyservicesinquiry/config/views.view.publications_search.yml
```diff
@@ -37,11 +37,11 @@ display:
       exposed_form:
         type: basic
         options:
-          submit_button: Search
+          submit_button: Apply
```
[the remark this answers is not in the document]
DuttonMa: Well yes, but the search form is not exposed so it won't actually have any visible effect - but have changed it anyway just to be safe

ooneill547 (approved):

PR #981, opened 2021-12-17
D8 un 840

soda736:
Help viewing documents link are added to every publication body field in NIPB. I am removing it from the body and adding it below the attachment field to match our other Unity sites.


web/sites/nipolicingboard/modules/custom/nipolicingboard_migrations/config/install/migrate_plus.migration.upgrade_d7_node_complete_publication_page.yml
```diff
@@ -100,6 +100,11 @@ process:
       plugin: site_files_filter
       from_ref: sites/nipb-drupal.nigov.net/files
       to_ref: files/nipolicingboard
+    -
+      plugin: str_replace
+      regex: true
+      search: /<p><a href="\/help-viewing-documents"><strong>Help viewing documents<\/strong><\/a><\/p>/
```
omahm, line 106: I'd double check that this markup is the same in all cases. I had an NIDirect ticket for the same type of thing and found around 4-5 different markup variants. Easiest way to check is to fire up Sequel Pro and search within the field data

soda736:
Thanks for pointing this out, I thought it was programmatically added to the body field in D7 but it wasn't. I found 3 different ways of it being added. There might be a tidier way to this but it works?

omahm:
> Thanks for pointing this out, I thought it was programmatically added to the body field in D7 but it wasn't. I found 3 different ways of it being added. There might be a tidier way to this but it works?

It's an improvement..  I knocked this together, it'll match any html element, multiple nested elements etc
`(<\w+>)*<a href=".*\/help-viewing-documents">(<\w+>)*Help viewing documents(<\/\w+>)*<\/a>(<\/\w+>)*`

This way it'll catch anything funky like: <div><p><a href="/help-viewing-documents"><strong>Help viewing documents</strong></a></p></div>

soda736:
Looks great! Thanks, works well :)

omahm (approved):

PR #1016, opened 2022-01-12
D8UN-794 Site level theming for revision comparison

ooneill547:


ooneill547:
> Looks good, just wondering if we can move some files to Unity Theme? There are a good few duplicates. e.g. block--system-main-block.html.twig and field--node--field-teaser--diff.html.twig. Save adding them in on per site basis everytime.

Yeah, there probably are some that are going to be the same in every site, but there were others that would have caused issues in one or two sites.

PR #1023, opened 2022-01-14
D8 un 891

soda736:


ooneill547 (approved):

web/sites/pbni/themes/pbni_theme/templates/views/views-view-unformatted--search-result-page.html.twig
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @see template_preprocess_views_view_unformatted()
+ */
+#}
+{% if title %}
+  <h3>{{ title }}</h3>
+{% endif %}
+{% for row in rows %}
+  {%
+    set row_classes = [
```
omahm, line 23: I'd move this outside the row loop

soda736: I've fixed these templates across Unity.

web/sites/default/settings.php
```diff
@@ -0,0 +1,792 @@
...
+ * For enhanced security, you may set this variable to the contents of a file
+ * outside your document root; you should also ensure that this file is not
+ * stored with backups of your database.
+ *
+ * Example:
+ * @code
+ *   $settings['hash_salt'] = file_get_contents('/home/example/salt.txt');
+ * @endcode
+ */
+$settings['hash_salt'] = 'dgVeucwlf22lXD_tZaEPZ4q_7Fbb0Q-OoDkqdv1FwTAUvJlSt6SiLCExq6bhmkkGj8mhlea3nA';
```
johangant: This value should, ideally, be kept as an environment variable.

web/sites/default/settings.php
```diff
@@ -0,0 +1,792 @@
...
+  'database' => 'drupal8',
+  'username' => 'drupal8',
+  'password' => 'drupal8',
+  'prefix' => '',
+  'host' => 'database',
+  'port' => '3306',
+  'namespace' => 'Drupal\\Core\\Database\\Driver\\mysql',
+  'driver' => 'mysql',
+);
+$settings['config_sync_directory'] = 'sites/default/files/config_aNJz9uTjokJEANnoHYx7eGWeylVqKYsxSDt-WT08GrvQiKiD7f5LDuN-kQeB5xmE7sHcqVwgDg/sync';
```
johangant: Looks like a default value from install. Should be an environment variable or at least something like `../config/<sitename>/sync`

web/sites/default/settings.php
```diff
@@ -0,0 +1,792 @@
...
+ * - Disabling JavaScript/CSS compression.
+ * - Rerouting outgoing emails.
+ *
+ * Keep this code block at the end of this file to take full effect.
+ */
+#
+# if (file_exists($app_root . '/' . $site_path . '/settings.local.php')) {
+#   include $app_root . '/' . $site_path . '/settings.local.php';
+# }
+$databases['default']['default'] = array (
```
johangant: The raw values here shouldn't be set like this. These settings can/should be set once and reference environment variables, which can be different from Lando to PSH, as required.

johangant (changes requested):
There are a number of settings and config values that should come from environment variables to avoid leaking the more sensitive things (site hash value) or just to avoid repetition (DB creds)

johangant (approved):
Unwanted settings files look gone, so all good :+1:

PR #1030, opened 2022-01-19
D8UN-859 Fiscal Council: Reconfigure and re theme front page

ooneill547:


soda736 (approved):
Looks good, do you need to add the new content types to the Workflow and Scheduled transitions?

ooneill547:
> Looks good, do you need to add the new content types to the Workflow and Scheduled transitions?

Oh, good point, I'll get that done now

PR #1033, opened 2022-01-19
D8UN-859 NI Fiscal Council: Reconfigure homepage graphs as an entityqueue

ooneill547:
Reconfigure homepage graphs as an entityqueue rather than a featured content list for ease of use. 
It is currently set to automatically add new features to the top of the list, but this can be turned off easily if the customer doesn't want it.


DuttonMa (approved):
Great work

PR #1043, opened 2022-01-26
D8 un 884

DuttonMa:
Implement Landing Page content type on Uregni and enable Layout Builder


web/sites/uregni/themes/uregni_theme/templates/block/block--views-block.html.twig
```diff
@@ -30,10 +30,12 @@
...
-{% endblock %}
+<div{{ attributes }}>
+  {% if label %}
+    <h2>{{ label }}</h2>
+  {% endif %}
+  {{ title_suffix }}
+  {% block content %}
+    {{ content }}
+  {% endblock %}
+</div>
```
DuttonMa, line 41: This may seem strange but adding the enclosing div and the {{ title_suffix }} ensured that the 'configure' icon appears correctly in the top right corner when adding a view block in layout builder

ooneill547 (approved):

PR #1084, opened 2022-02-11
D8UN-929 Invest NI: Configure basic page

ooneill547:


soda736 (approved):
LG 
There is no diff view mode but maybe be added in after.

PR #1140, opened 2022-02-21
Update drupal core to 9.2.13

ooneill547:


phpcs.sh
```diff
@@ -28,7 +28,7 @@ echo "Running coding standard checks in ${PHPCS_CHECK_DIR}"
 echo "Ignoring directories: ${IGNORE}"
 
 # Configure PHPCS.
-${PHPCS_PATH} --config-set installed_paths ${DRUPAL_DEPLOY_PATH}/vendor/drupal/coder/coder_sniffer
+#${PHPCS_PATH} --config-set installed_paths ${DRUPAL_DEPLOY_PATH}/vendor/drupal/coder/coder_sniffer
```
soda736, line 31: Just looking at NIDirect and this line has changed to 
`# Configure PHPCS.
${PHPCS_PATH} --config-set installed_paths ${DRUPAL_DEPLOY_PATH}/vendor/drupal/coder/coder_sniffer,${DRUPAL_DEPLOY_PATH}/vendor/slevomat/coding-standard`
it may be better trying this?

soda736 (commented):
I'm not sure if my comment will work but it has been changed on NIDirect recently. I had the issues in the Unity theme but it pulls in NIDirects phpcs.cs file.

phpcs.sh
ooneill547: Good catch, commenting it out was a quick fix to be able to test it, so I'll make this change now.

soda736 (approved):
LG

PR #1156, opened 2022-02-23
D8UN-915 Add social links to sites through the theme

ooneill547:


config/fiscalcommissionni/config/fiscalcommissionni_theme.settings.yml
```diff
@@ -10,3 +10,9 @@ favicon:
   use_default: 1
 ga_tracking_disabled: 1
 ga_tracking_disabled_eu: 0
+twitter_profile_url: 'https://twitter.com/fiscal_ni'
```
soda736, line 13: I don't ever see these changing but if these values are set in config then can they be changed in the UI? I think the purpose of these is to be able to add/remove/amend in the production environment without a deployment. Just wondering if these are set here and a change is made in production but not in this file then does this file config overwrite what is in production?
I think the config settings in config ignore deal with this without having to set the values in here?

soda736 (commented):
Also, should we add the block and theme settings to all sites regardless of whether they currently have social links?
This means that if that site needs a social link then everything is already set up and there is no need for a deployment to get the social link working.

config/fiscalcommissionni/config/fiscalcommissionni_theme.settings.yml
ooneill547: They still exported even with config ignore set. I think it just ignores it on import?

soda736 (approved):
LG

PR #1157, opened 2022-02-23
NIAO: Add production config for fastly etc

ooneill547:


config/niauditoffice/production/niauditoffice_theme.settings.yml
```diff
@@ -10,3 +10,9 @@ favicon:
   use_default: 1
 ga_tracking_disabled: 0
 ga_tracking_disabled_eu: 1
+twitter_profile_url: 'https://twitter.com/niauditoffice'
```
soda736, line 13: I don't ever see these changing but if these values are set in config then can they be changed in the UI? I think the purpose of these is to be able to add/remove/amend in the production environment without a deployment. Just wondering if these are set here and a change is made in production but not in this file then does this file config overwrite what is in production?
I think the config settings in config ignore deal with this without having to set the values in here?

johangant: I think @soda736 did some work around this whereby these were controllable from the admin UI. Or at least, the values defined in the UI took precedence over config.

The actual value here isn't a problem in my view as it's a public resource, but probably a conversation with @soda736 would be ideal to ensure this isn't running against the grain of a feature already added to Unity?

soda736 (approved):

PR #1172, opened 2022-03-02
D8UN-1008 PCSPS: Create pcsps migration module

ooneill547:


config/pcsps/config/core.extension.yml
```diff
@@ -75,6 +75,7 @@ module:
   metatag_open_graph: 0
   metatag_twitter_cards: 0
   metatag_views: 0
+  migrate_absolute_links: 0
```
soda736, line 78: We don't want to export this module.

soda736 (changes requested):
Martin said we want to get rid of alot of migration files. Mainly anything that has "field" in them as we never want to run those again. Check other migration modules and omit migration files from this module.

David-Corbett (approved):

soda736 (approved):
LG

PR #1196, opened 2022-03-11
Ag platform loop

soda736:


johangant (approved):
LGTM. The cron specs might be handy to spread out over the course of the sites / hour, otherwise each site kicks off a cron process at 10 mins past the hour, every hour. That could build up into a bit of pressure on underlying services... just depends on how much each site is doing etc.

PR #1202, opened 2022-03-11
D8UN-1040 CSCS: Create and theme basic page

ooneill547:


soda736 (approved):
This file web/sites/cscsreviewni/themes/cscsreviewni_theme/templates/navigation/menu--main--sidebar-second.html.twig is in the Unity theme now.

I'm not sure if you need this diff file anymore either? web/sites/cscsreviewni/themes/cscsreviewni_theme/templates/field/field--node--field-toc-enable--diff.html.twig

Otherwise all good.

PR #1262, opened 2022-04-22
CSCS: Post live config changes

ooneill547:


config/cscsreviewni/config/config_split.config_split.production.yml
```diff
@@ -9,7 +9,8 @@ folder: ../config/cscsreviewni/production
 module: {  }
 theme: {  }
 blacklist: {  }
```
soda736, line 11: Do we normally have fastly.settings in the production split?

ooneill547: yeah, I usually do anyway

soda736 (approved):
LG

PR #1275, opened 2022-04-28
D8 un 1064

DuttonMa:
Layout builder changes for Uregni. More changes to come but merging now to avoid branch falling too far behind development.


config/uregni/config/taxonomy.vocabulary.landing_pages_contact_card_colou.yml
```diff
@@ -0,0 +1,8 @@
+uuid: ac6f41f0-51fb-48c3-8b4f-20cdcf874fba
+langcode: en
+status: true
+dependencies: {  }
+name: 'Landing pages - contact card colours'
+vid: landing_pages_contact_card_colou
```
ooneill547, line 6: minor issue really, but the spelling 'error' could cause issues down the line

ooneill547 (approved):
Looks fine apart from the spelling error that it might be worth fixing

config/uregni/config/taxonomy.vocabulary.landing_pages_contact_card_colou.yml
DuttonMa: Good spot, but the name has actually hit a size limit so I can't add the missing character on the end !

PR #1335, opened 2022-05-16
D8UN-1137

soda736:
Adding a core patch https://www.drupal.org/project/drupal/issues/2906455 to enable a latest draft link for our workflow moderation views.


omahm (approved):
👍

PR #1336, opened 2022-05-16
D8 un 875

DuttonMa:
Remove contrib module cookie_content_blocker from composer as will be temporarily held in unity custom modules until the maintainers create a new release. (Otherwise it would be not be possible to upgrade PHP to 8.0 on Unity.)


omahm:
You'll need to update Circle CI as well: https://github.com/dof-dss/nicsdru_unity/blob/development/.circleci/config.yml#L132. etc

johangant (approved):
:+1:

PR #1410, opened 2022-06-28
D8 un 1156

KMDOFNI:


DuttonMa:
The pbni/settings.php file shouldn't be included in this PR as we don't add settings.php files to git

DuttonMa:
The placeholder images should be removed as we wouldn't add these to git

DuttonMa (approved):
All looks good now

PR #1472, opened 2022-08-03
D8UN-1183

AdamMcFet:


web/sites/pcsps/themes/pcsps_theme/src/js/pcsp.custom.js
```diff
@@ -0,0 +1,5 @@
+(function ($) {
```
DuttonMa, line 1: We should really convert this to Drupal.behaviours but I think that it makes sense to merge this and do that afterwards (as this solution works nicely)

DuttonMa (approved):
Looks good, great work !

PR #1503, opened 2022-09-05
D8UN-1076: oembed_providers module and config

johangant:
Extends https://github.com/dof-dss/nicsdru_unity/commit/2906ea7e54d85b0719e00af94ea645134a8f6c27


johangant:
NB: the deprecated code comes from an origins repo. I think that's been fixed upstream so a version bump should solve this.

PR #1588, opened 2022-11-08
D8 un 1495

soda736:


johangant (changes requested):
Careful with block adding config exports... there are bits of link checker config (https://github.com/dof-dss/nicsdru_unity/pull/1588/files#diff-980bbb0e9aee5160164d210ed7c86ca511d23657bdac8bb3730ee8d98952461c) as well as ultimate cron logging, blazy etc which don't seem relevant to the scope of the PR (add GA + GTM?)

These could well be benign and reflect the latest state of those sites/config, but do have a look over them and remove any you're not 100% sure about ahead of a merge just in case it introduces any unwanted changes.

soda736:
Thanks Johan, yes, just installing google tag and uninstalling GA on uregni. I've fixed all the other files.

johangant (approved):
Looks good now, cheers!

PR #1593, opened 2022-11-10
Make platform project id reference the same as the one on psh env vars

johangant:
Once deployed the PLATFORM_PROJECT_ID env var can be removed in Circle CI


DuttonMa (approved):
Good stuff !

PR #1597, opened 2022-11-11
D8UN-1488

DuttonMa:
Drupal core 9.4 upgrade and PHP 8.1 upgrade.
Lots of config changes here because when upgrading existing databases to Drupal 9.4 you have to run 'drush updb' and then export config so that (amongst other things) mysql is added to core.extension.yaml.


omahm (approved):
Quite the job, nice one!

PR #1830, opened 2023-08-30
node-sass / arm compilation support

johangant:
This seems to work for me, on an M2 chip:

Assuming these steps - based on an ad hoc script the team has used, run against the niauditoffice site theme as an example.

```
lando rebuild -y
cd project/sites/niauditoffice/themes/niauditoffice_theme
rm -rf node_modules
lando npm install 
lando npm install nicsdru_unity_theme 
lando npm run build
```

Note: it's intentionally forcing Lando to run the npm commands to isolate/insulate from the huge variations in host level tooling.


johangant:
NB: Also tested and works on x86_64 architecture chips

project/sites/niauditoffice/themes/niauditoffice_theme/package.json
```diff
@@ -90,5 +92,8 @@
     "> 1%",
     "last 2 versions",
     "ie >= 10"
-  ]
+  ],
+  "dependencies": {
+    "node-sass": "^9.0.0"
```
johangant, line 97: Hmm. Not sure how this bumped to v9... might be okay? It does seem to build... haven't tested with the full suite of commands though

johangant:
Tested across all site themes and on x86 + arm chips

johangant:
Going to look at integrating the build-themes.sh script into the CI pipeline tomorrow

build-themes.sh
```diff
@@ -0,0 +1,24 @@
...
+# Loop through each directory in the 'sites' directory
+for site in "$sites_dir"/*; do
+    site_name=$(basename "$site")
+    themes_dir="$site/themes/${site_name}_theme"
+
+    if [ -d "$themes_dir" ]; then
+        echo "Running npm install in $themes_dir"
+        cd "$themes_dir"
+
+        # if [ -d node_modules ]; then
```
johangant, line 16: @soda736 it would be nice if we didn't have to install/recompile all node modules each time this script is run, rather just pick up any new changes if the package-lock.json file changes. Do you think this is okay? @DuttonMa said you had some issues with this in the past... would be grateful if you could give me a few examples to help me see what might be going on.

johangant: Discussed today and seems like past problems with this stem from node_sass or other packages being built by developers using differing versions of node or build options.

Agreed to give this a try by relying on npm to honour the lock file versions and by standardising the version of node and compile flags within Lando and on Circle CI's container for the edge build, there should be very limited wiggle room for variations there.

ooneill547 (approved):

johangant:
Just noting here that with the introduction of a new env var to Lando, you'll need to rebuild your application in order for that to come into effect. `lando rebuild` should do it!

PR #1833, opened 2023-09-14
D8UN-2417

KMDOFNI:
updated twitter block with login text


David-Corbett (commented):
please make sure text strings pass through the Drupal t function

David-Corbett (approved):

PR #1864, opened 2023-11-06
Install legacy npm version using newer installer script

johangant:
- Means no more 20 seconds, followed by 60 second warning message on every application rebuild.
- Pruned away yarn. Can't see anything we use that exclusively relies on that and isn't compatible with npm


DuttonMa (approved):
Great work - good news !

PR #1868, opened 2023-11-08
Streamline npm install for edge build

johangant:


process.yml
```diff
@@ -0,0 +1,666 @@
+version: 2
+jobs:
+  build:
+    docker:
+    - image: thecodingmachine/php:8.1-v4-apache-node14
```
omahm, line 5: I'd replace all of these occurrences with a single docker image template e.g. <<: *docker_image so we can easily update the image in one place.

process.yml
```diff
@@ -0,0 +1,666 @@
...
+        PHP_INI_MAX_INPUT_TIME: 60
+    steps:
+    - attach_workspace:
+        at: ./
+    - run:
+        name: Deprecated code check
+        command: |
+          cd $PROJECT_ROOT
+          CHECK_DIRS="${PROJECT_ROOT}/web/modules/custom"
+          CHECK_DIRS="$CHECK_DIRS ${PROJECT_ROOT}/web/modules/origins"
```
omahm, line 94: I think we should also check the themes.

omahm (changes requested):
Some really nice checks in there, good job 👍 
Added in some minor comments.

process.yml
johangant: I forget why it was separated in the first instance, most likely quirks around the original method of adding npm. I've dropped the custom image and everything relies on the default one now

johangant:
That process.yml file shouldn't have been there. It's usually on disk after something like `circleci config process ~/.circleci/config.yml > process.yml` which can be used to compile the YML fragments into a full YML doc - CircleCI does this anyway each time it reads the config.yml file.

omahm:
> That process.yml file shouldn't have been there. It's usually on disk after something like `circleci config process ~/.circleci/config.yml > process.yml` which can be used to compile the YML fragments into a full YML doc - CircleCI does this anyway each time it reads the config.yml file.

I had pondered that TBH and thought you'd split the config across multiple files or something

omahm (approved):
LGTM

PR #1900, opened 2023-12-08
Wraith tests

DuttonMa:
Make screen capture work better and extend the range of pages tested on each site


johangant (approved):
Looks good, nice one

PR #1973, opened 2024-07-01
D8UN-2765

AdamMcFet:
- Building a custom module to programmatically input status label into a field on the consultation content type
- Cron task which runs to overwrite field with correct status when time passes
- Facet and block created to allow users to filter consultations by status
- Status field hidden from form view so authors can't edit it manually.

### DIFF
--- a/project/sites/uregni/modules/custom/uregni_consultations/uregni_consultations.module
+++ b/project/sites/uregni/modules/custom/uregni_consultations/uregni_consultations.module
@@ -5,13 +5,8 @@
  * Contains uregni_consultations.module.
  */
 
-use Drupal\Core\Database\Database;
-use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Entity\EntityInterface;
-use Drupal\node\NodeInterface;
-use Drupal\views\ViewExecutable;
 use Drupal\Core\Datetime\DrupalDateTime;
-use Drupal\Core\Render\Markup;
 
 /**
  * Implements hook_entity_presave().
@@ -19,7 +14,7 @@ use Drupal\Core\Render\Markup;
  */
 function uregni_consultations_entity_presave(EntityInterface $entity)
 {
-
+  // Only run this code on consultations
   if ($entity->bundle() == 'consultation') {
 
   // Get the current time to gauge consultation start and end dates off.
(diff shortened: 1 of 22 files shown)


DuttonMa (changes requested):
Great job overall, just a few tweaks needed to optimise the code in the module

project/sites/uregni/modules/custom/uregni_consultations/uregni_consultations.module
```diff
@@ -0,0 +1,74 @@
...
+ */
+function uregni_consultations_entity_presave(EntityInterface $entity)
+{
+  // Only run this code on consultations
+  if ($entity->bundle() == 'consultation') {
+
+    // Get the current time to gauge consultation start and end dates off.
+    $current_time = \Drupal::time()->getCurrentTime();
+
+    $entity = updateConsultationstatus($entity, $current_time);
```
DuttonMa, line 23: Think this should be updateConsultationStatus (a capital 'S' at the start of 'Status') - PHP is case sensitive

project/sites/uregni/themes/uregni_theme/uregni_theme.theme
```diff
@@ -39,8 +39,12 @@ function uregni_theme_preprocess_node(array &$variables) {
     if ($current_time > $consultation_start_date_timestamp && $current_time <= $consultation_end_date_timestamp) {
```
DuttonMa, line 39: I don't think that there should now be any logic in the theme, as you have populated the field in your module code. So you should be able to remove most of this code and just display the status.

DuttonMa (changes requested):
Just a small change in the module file and an idea about the theme

project/sites/uregni/themes/uregni_theme/uregni_theme.theme
```diff
@@ -38,18 +38,7 @@ function uregni_theme_preprocess_node(array &$variables) {
     // If the consultation started before and is ending after today's date then show the contact information.
```
DuttonMa, line 38: I think that you can still improve this theme code. I can see that you need to set $variables['consultation_in_progress'] but your new current status field should be available in the $variables array so you don't need the date/time logic.
Try this code to see the value in the current status field:

$status = $variables['node']->get('field_current_status')->value;
\Drupal::logger('uregni_theme')->error("Status is $status");

... so you should just be able to set $variables['consultation_in_progress'] based on the value that you find in the current status field shouldn't you ?

DuttonMa (changes requested):
Have left a comment about the theme.

I also noticed that when I load up a dump of the latest production Uregni database locally, if I then try a config import for Uregni it complains that there is no field storage for the current status field - so maybe you need to go back over your config and make sure that you have exported everything ?

DuttonMa (approved):
Great work !

PR #1992, opened 2024-08-12
D8UN-2826

AdamMcFet:
- Removing background styling for UREGNI social icons
- Fixes issue caused when hovering icon
- Had to increase node-sass to 9.0.0 for NPM to work (9 is used on a lot of the newer unity 3 sites)


AdamMcFet:
> Did updating node sass cause any issues?

Shouldn't cause any issues, css built correctly and site all looked fine locally. I use 9.0.0 on most of the Unity 3 sites.

ooneill547 (approved):

PR #2201, opened 2026-06-23
D8 un 3469

johangant:


johangant:
Sorry, ridiculously large PR given the number of files changed. Probably best if it's evaluated this way:

1. Check out the branch locally on DDEV. `ddev restart` to update with the new service container. Then reindex and check behaviour.
2. Spin up an Upsun dev env and evaluate there too.

soda736 (approved):

